PR #1, opened 2016-11-10
Add SCR features needed to test HBWIF

jwright6323:
Adds some singletons to SCR.scala and a PutSeqDriver utility to be able to drive sequences of SCR configurations for TL unit tests. Needed for HBWIF.

### DIFF
--- a/src/main/scala/HeaderEnum.scala
+++ b/src/main/scala/HeaderEnum.scala
@@ -1,7 +1,7 @@
 package testchipip
 
 import Chisel._
-import scala.collection.mutable.HashMap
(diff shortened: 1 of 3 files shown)


colinschmidt:
This is just a convenience function? There are two separate things in this commit right? The SCR changes adding that hashmap and then the widget that puts a seq of tuples. Perhaps they should be separate commits and a usecase would be good to know. I would have thought the second thing would already exist for the other unit testers.

zhemao:
No, the SCR changes add stuff to a global object, and the global object contains a function to dump it all out. I think it's fine to put them together, but I'm a bit wary about the global object. Not sure if there's a better way, though.

PR #2, opened 2016-11-24
Use new IO style to fix compilation error

grebe:
Use the new IO style. If you don't like this (because e.g. you're on an old version of chisel3) this should be written in compatibility mode.


zhemao:
I guess our use of Chisel vs. chisel3 is inconsistent. Our working servers are down for maintenance right now, but I'll try to get everything switched over to chisel3 once they're back up.

grebe:
Superseded by #3

PR #3, opened 2016-12-07
Switch imports from Chisel._ to chisel3._

grebe:
Per comment in #2, I thought I'd try switching all `Chisel._` imports to `chisel3._`. It works for me in the limited way I use the SCR.

The only real hiccup I found is that rocket's util package has a namespace collision with chisel3.util (at least, if you `import chisel3._`). This can be solved by importing util before chisel3, but I think it is more clear if you import `_root_.util`. Rocket should probably change package util to have a better name.


zhemao:
I would actually prefer importing util._ before chisel3._ instead of importing _root_.util. Otherwise this looks alright to me.

zhemao:
This supersedes #2 right?

grebe:
Correct, I'll close #2. I didn't want to close it in case you weren't ready to port everything over to chisel3 yet.

chick:
@zhemao : Just curious, why do you prefer the order specific import, isn't that more brittle.

zhemao:
Okay, so one issue is that this would require us to upgrade the rocket-chip submodule in project-template and fpga-zynq to get the updated chisel submodule. But that would also require us to upgrade everything to TileLink2, which is a larger undertaking, but one we should do anyway.

Given that, I think we should do it all in one go.

zhemao:
@chick I guess I don't have a good reason other than `_root_` looking ugly. I suppose it's fine to keep it as is.

grebe:
Our stopgap to that particular problem has been to fork rocketchip and just bump chisel and firrtl.

I agree that upgrading everything to TileLink2 is ultimately the way to go and would certainly be happy to help as much as my rocket knowledge allows.

grebe:
@zhemao should I revert? Or keep the current?

src/main/scala/ChannelMapper.scala
```diff
@@ -10,10 +11,10 @@ import cde.Parameters
  * channel are contiguous.
  */
 class ChannelAddressMapper(n: Int)(implicit p: Parameters) extends Module {
-  val io = new Bundle {
+  val io = IO(new Bundle {
     val in  = Vec(n, new ClientUncachedTileLinkIO).flip
     val out = Vec(n, new ClientUncachedTileLinkIO)
```
jackkoenig: Shouldn't there be directions here? I'm actually surprised it works without it.

grebe: ClientUncachedTileLinkIO is defined in the Chisel emulation layer, so all the ports should have OUTPUT or INPUT (which presumably calls Output and Input?). This is actually something I'm not sure I understand- what are you meant to do when your port is a bundle (i.e. something that doesn't just have one direction? Should it be wrapped in Output()?

zhemao:
I think it's fine to keep this PR up. I'll fold the changes into an eventual TL2 revamp.

src/main/scala/ChannelMapper.scala
jackkoenig: Oh duh, my bad. Yeah if you were to wrap in Input or Output it would coerce all of the directions.

Instead of `.flip` you should wrap in `Flipped(...)` though

src/main/scala/Adapter.scala
```diff
@@ -14,6 +14,8 @@ import rocketchip._
 import rocket.XLen
 import cde.{Parameters, Field}
 
+import chisel3.core.ExplicitCompileOptions.NotStrict
```
grebe: @jackkoenig does it make sense that I need this? It says `Not bound to synthesizable node, currently only Type description` on line 210.

grebe: Oh, I see. Is it true that the Chisel compatibility layer doesn't call IO()? The io is being instantiated by a base class in rocket.

davidbiancolin:
@zhemao Just for the record, i've taken the time to bump a branch of fpga-zynq to master of rocketchip, in which the serial adapter still speaks TL1, but is connected to a TL2 periphery/L2 crossbar. You might also consider folding some of the changes in the fpga_zynq_bump of testchipip branch when you decide to move foward.

zhemao:
Oh, right, I guess you can use the legacy adapter to keep the serial port in TileLink1. I'll try doing that.

edwardcwang:
@zhemao Do you know when the TL2 revamp would happen by any chance?

PR #8, opened 2017-08-04
Allow blockdevice databits to be overriden

colinschmidt:
This enables adapting to different system buses

Can use blockdevice with hwacha configurations this way.


zhemao:
Have you actually tested this with a setting other than 64? I tried doing this before, but getting the software model to work with variable widths was kind of tricky.

colinschmidt:
Yeah you're right I haven't gotten the software model to work yet. Looking at that soon though.

PR #11, opened 2017-11-17
Remove IntIdentityNode; Use TL := operator thingy.

ccelio:
I didn't try this in vivo, but it at least compiles using the latest rocket-chip.


ccelio:
Alrighty, SimSerial shown to work on BOOM running Hello World.

zhemao:
Ugh. Do I need to update rocket-chip in our project-template in order to make this compile?

ccelio:
? 

This is an update to testchipip so testchipip works with newer rocket-chips. Naturally any upstream repo that uses testchipip will need to also bump rocket-chip to match.

zhemao:
I just wasn't expecting to have to update rocket-chip at this juncture. We have a bunch of other submodules that also need to be kept in sync.

Issue #13, opened 2018-02-02
Failed to generate emulator with errors


cokeco:
Hi, 
I clone the project-template project and desire to run it.
However, there are errors at build emulator stage.

[error] /RocketPlatform/testchipip/src/main/scala/Generator.scala:16: type mismatch;
[error]  found   : freechips.rocketchip.util.ParsedInputNames
[error]  required: Seq[String]
[error]   lazy val config = getConfig(names)
[error]                               ^
[error] /RocketPlatform/testchipip/src/main/scala/Generator.scala:18: value root is not a member of object freechips.rocketchip.config.Parameters
[error]   lazy val params = Parameters.root(world)
[error]                                ^
 [error] two errors found

It seems Generator.scala( in src/main/scala) needs to revise.
I am not sure that does it cause by using the latest rocket-chip version?

colinschmidt:
Closed by #14

PR #14, opened 2018-02-02
fix for deprecated Parameters.root

RaulGoal:
Revise src/main/scala/Generator.scala : address issue #13

### DIFF
--- a/src/main/scala/Generator.scala
+++ b/src/main/scala/Generator.scala
@@ -1,7 +1,7 @@
 package testchipip
 
 import chisel3.{Module, Driver}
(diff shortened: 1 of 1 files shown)


src/main/scala/Generator.scala
```diff
@@ -1,7 +1,7 @@
 package testchipip
 
 import chisel3.{Module, Driver}
-import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.config._
```
colinschmidt: With only 2 imports can we use the `.{Parameters, Config}` syntax

colinschmidt:
Thanks

PR #15, opened 2018-03-22
Tl2 switcher

zhemao:


src/main/scala/Switcher.scala
```diff
@@ -0,0 +1,147 @@
...
+    val sel = Input(UInt(log2Ceil(n).W))
+  })
+
+  io.out.zipWithIndex.foreach { case (out, i) =>
+    val selected = io.sel === i.U
+    out.a.valid := io.in.a.valid && selected
+    out.a.bits := io.in.a.bits
+    out.d.ready := io.in.d.ready && selected
+
+    out.b.ready := false.B
```
colinschmidt: Can we have an assertion on io.in.b.valid (and the other unused channels)?

src/main/scala/Switcher.scala
```diff
@@ -0,0 +1,147 @@
...
+    val a = Wire(Decoupled(new TLBundleA(params)))
+    a.valid := in.a.valid
+    a.bits := in.a.bits
+    a.bits.source := Cat(in.a.bits.source, i.U(log2Ceil(n).W))
+    in.a.ready := a.ready
+    a
+  }
+
+  TLArbiter.robin(edge, io.out.a, inA:_*)
+  io.out.d.ready := false.B
```
colinschmidt: Can you move this to above the zip so its clear this is a different kind of default

zhemao: Ah, right

src/main/scala/Switcher.scala
zhemao: It will just freeze up if it actually tries to send, won't it? Since the ready is false.

colinschmidt: Yeah I just want the assertion for simulation. Its easy to debug an assertion than a timeout.

colinschmidt:
I CR'd this in person

PR #24, opened 2018-10-23
error-device: use current rocket-chip HEAD tile-link error parameters

stettberger:
The current freechips/rocket-core has no ErrorParams class, but it was renamed to DevNullParams.


jwright6323:
This is going in with #31. Thanks for the PR!

PR #25, opened 2018-10-31
Import rocket's GeneratorApp directly instead of copying

edwardcwang:
It's currently out of date and can cause casting errors

### DIFF
--- a/src/main/scala/Generator.scala
+++ b/src/main/scala/Generator.scala
@@ -2,43 +2,12 @@ package testchipip
 
 import chisel3.{Module, Driver}
 import freechips.rocketchip.config.{Parameters, Config}
-import freechips.rocketchip.util.{HasGeneratorUtilities, ParsedInputNames}
+import freechips.rocketchip.util.{GeneratorApp, HasGeneratorUtilities, ParsedInputNames}
 import java.io.{File, FileWriter}
 import firrtl.annotations.JsonProtocol
 
-trait GeneratorApp extends App with HasGeneratorUtilities {
-  lazy val names = ParsedInputNames(
-    targetDir = args(0),
-    topModuleProject = args(1),
-    topModuleClass = args(2),
-    configProject = args(3),
-    configs = args(4))
-
-  lazy val config = getConfig(names.fullConfigClasses)
-  lazy val params = config.toInstance
-  lazy val circuit = Driver.elaborate(() =>
-      Class.forName(names.fullTopModuleClass)
-        .getConstructor(classOf[Parameters])
-        .newInstance(params)
-        .asInstanceOf[Module])
-
-  lazy val longName = names.topModuleProject + "." +
-                 names.topModuleClass + "." +
-                 names.configs
-
-  def generateFirrtl =
-    Driver.dumpFirrtl(circuit,
-      Some(new File(names.targetDir, s"$longName.fir")))
-
-  def generateAnno {
-    val annoFile = new File(names.targetDir, s"$longName.anno")
-    val afw = new FileWriter(annoFile)
-    afw.write(JsonProtocol.serialize(circuit.annotations.map(_.toFirrtl)))
-    afw.close()
-  }
-}
-
 object Generator extends GeneratorApp {
+  val longName = names.topModuleProject + "." + names.topModuleClass + "." + names.configs
   generateFirrtl
   generateAnno
 }


src/main/scala/Generator.scala
```diff
@@ -5,38 +5,7 @@ import freechips.rocketchip.config.{Parameters, Config}
...
-  def generateAnno {
-    val annoFile = new File(names.targetDir, s"$longName.anno")
-    val afw = new FileWriter(annoFile)
-    afw.write(JsonProtocol.serialize(circuit.annotations.map(_.toFirrtl)))
-    afw.close()
-  }
-}
+import freechips.rocketchip.util.GeneratorApp
 
 object Generator extends GeneratorApp {
```
jwright6323: Does this need to be ```object Generator extends freechips.rocketchip.util.GeneratorApp {```?

jwright6323:
Do we need:
https://github.com/ucb-bar/testchipip/commit/4aedbaf8c32805350f59ecd8b82bd833810fec80#diff-8abc7f24bbe4cfc68dcbdcd13daa9ce4R6

edwardcwang:
Done.

src/main/scala/Generator.scala
edwardcwang: I think once it's been imported we don't need to refer to the fully-qualified/long name.

edwardcwang:
Requires a diff in project-template, will PR later with a bump to testchipip

```diff
diff --git a/src/main/scala/example/TestHarness.scala b/src/main/scala/example/TestHarness.scala
index e67a189..d97b89e 100644
--- a/src/main/scala/example/TestHarness.scala
+++ b/src/main/scala/example/TestHarness.scala
@@ -3,7 +3,7 @@ package example
 import chisel3._
 import freechips.rocketchip.diplomacy.LazyModule
 import freechips.rocketchip.config.{Field, Parameters}
-import testchipip.GeneratorApp
+import freechips.rocketchip.util.GeneratorApp
```

edwardcwang:
@zhemao It seems to work (elaborates and builds) with the above changes to project-template. Tested as far as building fesvr (but this fix shouldn't affect building fesvr anyway, since it's a scala-land change)

zhemao:
Is there a reason you omitted `topModuleClass` in `longName`? I would rather the makefile not have to be changed. Sometimes we might want to synthesize different top-level modules in a package.

edwardcwang:
@zhemao They must have changed it in upstream rocket-chip - the `longname` in this patch is copied directly from upstream rocket-chip (in fact the `Generator` class will not compile without it).

zhemao:
The makefile in project-template is different from the one in rocket-chip. That's deliberate. I know you need to define longName. I'm saying you should define it to the same as it was before.

edwardcwang:
I see; trying a fix.

edwardcwang:
New diff in project-template:

```diff
diff --git a/Makefrag b/Makefrag
index 4a6be9f..62f6efc 100644
--- a/Makefrag
+++ b/Makefrag
@@ -25,7 +25,7 @@ include $(testchip_dir)/Makefrag
 CHISEL_ARGS ?=
 
 FIRRTL_FILE=$(build_dir)/$(PROJECT).$(MODEL).$(CONFIG).fir
-ANNO_FILE=$(build_dir)/$(PROJECT).$(MODEL).$(CONFIG).anno
+ANNO_FILE=$(build_dir)/$(PROJECT).$(MODEL).$(CONFIG).anno.json
 VERILOG_FILE=$(build_dir)/$(PROJECT).$(MODEL).$(CONFIG).v
 
 $(FIRRTL_FILE) $(ANNO_FILE): $(SCALA_SOURCES) $(bootrom_img) $(FIRRTL_JAR)
diff --git a/src/main/scala/example/TestHarness.scala b/src/main/scala/example/TestHarness.scala
index e67a189..20b7dc0 100644
--- a/src/main/scala/example/TestHarness.scala
+++ b/src/main/scala/example/TestHarness.scala
@@ -3,7 +3,7 @@ package example
 import chisel3._
 import freechips.rocketchip.diplomacy.LazyModule
 import freechips.rocketchip.config.{Field, Parameters}
-import testchipip.GeneratorApp
+import freechips.rocketchip.util.GeneratorApp
 
 case object BuildTop extends Field[(Clock, Bool, Parameters) => ExampleTopModule[ExampleTop]]
 
@@ -21,6 +21,7 @@ class TestHarness(implicit val p: Parameters) extends Module {
 }
 
 object Generator extends GeneratorApp {
+  val longName = names.topModuleProject + "." + names.topModuleClass + "." + names.configs
   generateFirrtl
   generateAnno
 }
```

edwardcwang:
@zhemao I added the model back like the old code and it appears to work with the corresponding change in project-template (plus an update to the annotations format).

PR #27, opened 2018-12-11
Fix PRIx64 problem with certain gcc versions

jwright6323:
This fixes a problem I'm seeing with gcc versions on the BWRC machines and inttypes.h.

### DIFF
--- a/csrc/blkdev.cc
+++ b/csrc/blkdev.cc
@@ -1,7 +1,7 @@
 #include "blkdev.h"
-#include <stdlib.h>
-#include <string.h>
-#include <inttypes.h>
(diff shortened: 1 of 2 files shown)


zhemao:
Well that's odd.

If that's the case, you should also change the other standard library includes for consistency.

    #include <cstdlib>
    #include <cstring>

And change the ones in blkdev.h too.

    #include <cstdlib>
    #include <cstdint>

jwright6323:
Good idea, I'll do that.

jwright6323:
@zhemao Looks good to me now. Can you check on your env and merge if it works?

PR #28, opened 2019-01-18
Add clock utilities

jwright6323:
This adds some clock utilities that previously lived in eagle-pll, with some added withGatedClock(AndReset) objects and removed tech-specific implementations of the clock cells.


jwright6323:
This passes unit tests and doesn't touch existing code, so I'm going to go ahead and bring this in.

PR #29, opened 2019-01-18
make manager >= client sourceBits.

harrisonliew:
Previous assertion was impossible to meet. This will still work assuming you don't create a full loop both on chip and off.


jwright6323:
This is fine, CR'ed with @colinschmidt in person.

colinschmidt:
I think it would be cool if @zhemao looked at this as well but I think we are reasonably confident. Some runtime testing would be good but I think @harrisonliew is trying that out now.

PR #31, opened 2019-02-12
Make publishable

grebe:
Restructure so testchipip can be used as a published artifact.


bootrom
```diff
@@ -0,0 +1 @@
+src/main/resources/testchipip/bootrom/
```
zhemao: Is this a symlink now? Is it still necessary to have this?

grebe: They aren't necessary for my project-template PR, but I don't know how everyone else uses testchipip. I thought it best to make the least intrusive change.

PR #33, opened 2019-02-20
[WIP] TSI Host Widget

abejgonzalez:
A first implementation of a TSI Host Widget. Takes TSI commands through MMIO, converts them into a decoupled interface, converts them to TL requests, and sends the requests over a Serial interface. Reuses the SerialAdapter within the repo. Comes with a basic unit test to test the TSI read requests by bypassing the MMIO.

ToDo:
- [x] Create "full system" unit test to test the widget (aka test the MMIO and write TSI commands)
- [x] Basic unit tests
- [x] First implementation of widget

Questions:
1. Instead of creating a `fesvr` folder, would it be better to consolidate sources into a single file?
2. Is there a need for a write TSI unit test if I create a full system unit test?
3. Any particular defaults wanted? (Default base, etc)


jwright6323:
Yay!

I'll look at this later but I think it might more accurately be titled "TSI Host Widget" since fesvr lives in software land.

src/main/scala/Configs.scala
```diff
@@ -4,6 +4,7 @@ import chisel3._
 import freechips.rocketchip.system.BaseConfig
 import freechips.rocketchip.config.{Parameters, Config}
 import freechips.rocketchip.unittest.UnitTests
+import freechips.rocketchip.diplomacy._
```
jwright6323: Why do we need `diplomacy` in Configs? I don't see any other changes in this file so this is odd to me.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+  // this should communicate over MMIO to the core (tx and rx) and
+  // should covert the communication from TL (aka TL -> Parse -> MMIO, MMIO -> Parse -> TL)
+  val mmio: TLTSIHostMMIO = if (p(PeripheryTSIHostKey).bypassMMIO) { null } else { LazyModule(new TLTSIHostMMIO(beatBytes)) }
+  // reuse module from Howie to take in a decoupled io tsi and convert to a TL stream
+  val parseToTL = LazyModule(new SerialAdapter)
+  // This converts the TL signals given by the parser into a decoupled stream
+  val serdes = LazyModule(new TLSerdesser(
+        w = p(PeripheryTSIHostKey).serialIfWidth,
+        clientParams = TLClientParameters(
+          name = "tl-ser",
```
jwright6323: Not crazy about "tl-ser", how about "tl-tsi-host" or something similar?

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+class TLTSIHostWidget(val beatBytes: Int)(implicit p: Parameters)
+  extends LazyModule
+{
+  // this should communicate over MMIO to the core (tx and rx) and
+  // should covert the communication from TL (aka TL -> Parse -> MMIO, MMIO -> Parse -> TL)
+  val mmio: TLTSIHostMMIO = if (p(PeripheryTSIHostKey).bypassMMIO) { null } else { LazyModule(new TLTSIHostMMIO(beatBytes)) }
+  // reuse module from Howie to take in a decoupled io tsi and convert to a TL stream
+  val parseToTL = LazyModule(new SerialAdapter)
+  // This converts the TL signals given by the parser into a decoupled stream
+  val serdes = LazyModule(new TLSerdesser(
```
jwright6323: There are some assertions here that are likely to cause some issues: https://github.com/ucb-bar/testchipip/blob/4775caf9f23826fed9e2400c48ca0cfb88f9eb8f/src/main/scala/Serdes.scala#L554-L558
This is designed to prevent you from messing up the id bitwidths on-chip, but the requirements may be reversed on the host side.

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
```diff
@@ -0,0 +1,172 @@
...
+  // implementation of the module
+  lazy val module = new LazyModuleImp(this) {
+    // i/o to connect to the unit test interface
+    val io = IO(new Bundle with UnitTestIO)
+
+    // TSI cmd of the form (cmd, lowAddr, highAddr, lowLen, highLen)
+    // note that len is actually the len - 1 (e.g. len of 4 is really 5)
+    // and len is in increments of 32b
+    // O is read, 1 is write
+    val inputTsiSeq = Seq(Seq(0,  0x0, 0, 0x0, 0),
```
jwright6323: We probably want a helper object here to generate commands that are of some TSICommand class type. Not exactly this, but something like:

`val inputTsiSeq = Seq(TSIWrite(0x80000000, 0xdeadbeef), TSIRead(0x80000000), TSIWrite(0x80000008, 0xffff0000))`

It may even be worth making your own `TSIFuzzer`

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
```diff
@@ -0,0 +1,172 @@
...
+    targetRomData.flatMap(
+      data => (0 until 8).map(i => ((data >> (i * 8)) & 0xff).toByte)),
+    beatBytes = beatBytes))
+
+  // TLFragmenter splits up the multi beat packet into single beats that the left hand side can take
+  // (of size beatBytes).
+  targetRom.node := TLFragmenter(beatBytes, lineBytes) := TLBuffer() := targetSerdes.clientNode
+
+  // fuzzer that lives in target-land that sends random requests to the ram on the host
+  val targetFuzzer = LazyModule(new TLFuzzer(
```
jwright6323: This is going to make it hard to write a test that will deterministically pass, because this can arbitrarily overwrite the data you are sending on TSI without your tracker knowing about it.

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
```diff
@@ -0,0 +1,172 @@
...
+      supportsPutPartial = TransferSizes(1, lineBytes)),
+    beatBytes = beatBytes))
+
+  // rom living in target-land. connects to the client of the serdes
+  val targetRomData = Seq(
+    BigInt("7766554433221100", 16),
+    BigInt("FFEEDDCCBBAA9988", 16),
+    BigInt("0123456789ABCDEF", 16),
+    BigInt("FEDCBA9876543210", 16),
+    BigInt("76543210FDECBA98", 16))
```
jwright6323: Can this be randomly generated instead?

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
```diff
@@ -0,0 +1,172 @@
...
+                          Seq(0,  0xC, 0, 0x0, 0),
+                          Seq(0, 0x10, 0, 0x0, 0),
+                          Seq(0, 0x20, 0, 0x1, 0))
+
+    // 32b sequences expected from the TSI response
+    val outputTsiData = Seq(Seq(BigInt("33221100", 16)),
+                            Seq(BigInt("77665544", 16)),
+                            Seq(BigInt("FFEEDDCC", 16)),
+                            Seq(BigInt("89ABCDEF", 16)),
+                            Seq(BigInt("FDECBA98", 16), BigInt("76543210", 16)))
```
jwright6323: This seems fragile, this should be some function that derives the responses from the inputTsiSeq and the known state, rather than a hardcoded thing. Having the TSICommand objects like mentioned above might make this easier.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/**
+ * I/O to the outside world. This is the stream data out of the TSIHostWidget.
+ *
+ * @param w width in bits of connection to outside world
+ */
+class TSIHostWidgetIO(val w: Int)(implicit val p: Parameters) extends Bundle {
+  val serial = new SerialIO(w)
+  val debug: SerialIO = if(p(PeripheryTSIHostKey).bypassMMIO) { new SerialIO(w) } else { null } // unconnected in normal operation
```
jwright6323: This might not be needed due to comment above, but some feedback here:

`null` is a big scala no-no. Use an `Option`:
`val debug: Option[SerialIO] = p(PeripheryTSIHostKey).option(new SerialIO(w))`

You'll need to call `.get` on places where you use `debug`

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+ * Rx is TL -> Deserializer -> Queue -> MMIO
+ *
+ * @param fesvrParams parameter object
+ */
+class TLTSIHostWidget(val beatBytes: Int)(implicit p: Parameters)
+  extends LazyModule
+{
+  // this should communicate over MMIO to the core (tx and rx) and
+  // should covert the communication from TL (aka TL -> Parse -> MMIO, MMIO -> Parse -> TL)
+  val mmio: TLTSIHostMMIO = if (p(PeripheryTSIHostKey).bypassMMIO) { null } else { LazyModule(new TLTSIHostMMIO(beatBytes)) }
```
jwright6323: See `null` and `Option` comment above

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+  mmioRegWidth: Int = 32,
+  serialIfWidth: Int = 32,
+  txQueueEntries: Int = 16,
+  rxQueueEntries: Int = 16,
+  baseAddress: BigInt = BigInt(0x10017000),
+  tlAddressSet: AddressSet = AddressSet(0, BigInt("FFFFFFFF", 16)),
+  tlRegionType: RegionType.T = RegionType.UNCACHED,
+  tlNumTransactions: Int = 1,
+  tlLineBytes: Int = 64,
+  bypassMMIO: Boolean = false
```
jwright6323: Rather than use a bypass can you use a `BundleBridge` and a `TLPatternPusher` in your test to perform the TL transactions you need?

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+)
```
jwright6323: I think this case class is missing the manager and client params fields. Where do those get set?

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,171 @@
...
+ * Top level class that uses a register router to connect MMIO to the core. This instantes
+ * the implementation module and bundle
+ *
+ * @param beatBytes amount of bytes to send per beat
+ */
+class TLTSIHostMMIO(val beatBytesIn: Int)(implicit p: Parameters)
+  extends TLRegisterRouter(
+    base = p(PeripheryTSIHostKey).baseAddress,
+    devname = "fesvr-mmio",
+    devcompat = Seq("ucbbar,fesvr-widget"),
```
jwright6323: `devcompat = Seq("ucbbar,tsi-host"),`

src/main/scala/Configs.scala
abejgonzalez: Will go ahead and delete. This was left from a previous commit.

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: Is there an example of `BundleBridge` used somewhere else? The goal of the bypass was to be able to still send TSI requests without going through the MMIO. The `SerialAdapter` would then get this and send/convert the TSI into TL. If I understand `TLPatternPusher` correctly, this is just used to create TL request sequence. That is already handled by the `SerialAdapter`.

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: How about, "tl-tsi-host-ser"? I think that is descriptive enough

jwright6323: Well I guess I don't really like "ser" because it's both "ser" and "des" "tl-tsi-host-serdes" is probably fine.

src/main/scala/tsiHost/TsiHostWidget.scala
jwright6323: What I meant was that you could generate the transactions that you'd send to the MMIO. Otherwise I'm not a big fan of the boolean to add ios for simulation only. I'd say at that point just unit test the underlying (non-diplomatic) widgets and we'll have a larger integration test that makes sure the diplomacy stuff works. @colinschmidt ideas here?

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: Okay. I think that is good. Ill use that

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: Offline Conversation: For now, create an inner block with a SerialIO interface to connect to (aka split the Frontend MMIO from the Backend Adapter + TLSerdesser). Unit test that block. Wait to work on adding a TLPatternPusher to the MMIO end.

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
abejgonzalez: Other than this widget will there be a need to create a TSIFuzzer? For now, I think just a helper factory would be good.

jwright6323: Maybe not, let's see what it looks like after you make some changes.

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
abejgonzalez: In my UnitTest, shouldn't the TLFuzzer interact with the Host RAM and thus would not affect the Target RAM. So the TSI commands will not collide with the TLFuzzer transactions?

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: Would it be acceptable to have a boolean to swap the requirements? Or is there a preferred method to do this?

abejgonzalez: Offline: Disable when instantiated on host.

src/main/scala/Serdes.scala
```diff
@@ -551,11 +552,18 @@ class TLSerdesser(
...
+      //This assumes that on chip, the client node can drive the manager
+      require (client_tl.params.sourceBits  <= manager_tl.params.sourceBits)
+      require (client_tl.params.addressBits == manager_tl.params.addressBits)
+      require (client_tl.params.dataBits    == manager_tl.params.dataBits)
+    }
+
+    println(s"${clientParams.name}: Client:  (addressBits, dataBits, sourceBits, sinkBits, sizeBits)")
+    println(s"                               (${client_tl.params.addressBits}, ${client_tl.params.dataBits}, ${client_tl.params.sourceBits}, ${client_tl.params.sinkBits}, ${client_tl.params.sizeBits})")
+    println(s"${clientParams.name}: Manager: (addressBits, dataBits, sourceBits, sinkBits, sizeBits)")
+    println(s"                               (${manager_tl.params.addressBits}, ${manager_tl.params.dataBits}, ${manager_tl.params.sourceBits}, ${manager_tl.params.sinkBits}, ${manager_tl.params.sizeBits})")
```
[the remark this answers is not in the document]
abejgonzalez: These will be removed/changed in the final merge of the widget. These were just used for debugging.

src/main/scala/Serdes.scala
```diff
@@ -107,7 +107,7 @@ class StreamWidener(inW: Int, outW: Int) extends Module {
   val inBeats = outW / inW
 
   val data = Reg(Vec(inBeats, UInt(inW.W)))
-  val keep = RegInit(Vec(Seq.fill(inBeats)(0.U(inBytes.W))))
+  val keep = RegInit(VecInit(Seq.fill(inBeats)(0.U(inBytes.W))))
```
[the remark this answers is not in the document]
abejgonzalez: I think this was just to avoid the deprecation warnings that Chisel3 was giving.

src/main/scala/tsiHost/TsiHostPeriphery.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import chisel3._
+import chisel3.util._
+
+import freechips.rocketchip.config.{Parameters, Field}
+import freechips.rocketchip.subsystem._
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.util._
+import freechips.rocketchip.amba.axi4._
```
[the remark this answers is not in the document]
abejgonzalez: Currently, no. I will clean these up in the final PR push.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,315 @@
...
+    supportsGet = TransferSizes(1, 64),
+    supportsPutFull = TransferSizes(1, 64),
+    supportsPutPartial = TransferSizes(1, 64))
+)
+
+
+/**
+ * TSI Host parameter class
+ *
+ * @param mmioRegWidth size of the MMIO data being sent back and forth in bits (connects to SerialAdapter)
```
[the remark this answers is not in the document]
abejgonzalez: Not from what I did, you have to specify the width in the `RegField` entry. This is also used to determine the interface width from the frontend (aka the mmio regs) to the backend (the serial adapter)

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,315 @@
...
+  rxQueueEntries: Int = 16,
+  baseAddress: BigInt = BigInt(0x10017000),
+  serdesParams: TSIHostSerdesParams = TSIHostSerdesParams()
+)
+
+/**
+ * Offsets for the MMIO communication queues (base + offset to get the proper address)
+ */
+object TSIHostWidgetCtrlRegs {
+  val txQueueOffset = 0x00 // note: these assume mmioRegWidth = 32b
```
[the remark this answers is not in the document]
abejgonzalez: Currently this is `required` indirectly in the backend by the statement `require(params.mmioRegWidth == SERIAL_IF_WIDTH)` (where the SERIAL_IF_WIDTH) is 32. I can add another require in the frontend for this also if you'd like.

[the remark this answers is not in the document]
jwright6323: IMO we shouldn't expose mmioRegWidth as a param at all.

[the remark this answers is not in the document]
abejgonzalez: This depends on the future of the widget and FESVR. If we want more BW we can send a 64b data value through the system. This would give a little bit more speed but would require changes to this widget, and to the serialadapter. If you don't think that'll happen anytime soon, we can just hardcode that.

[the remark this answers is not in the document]
abejgonzalez: Note: There can be some confusion between `serialIfWidth` and `SERIAL_IF_WIDTH`. The former is the interface with going out/into the serdes while the latter is the interface with from the MMIO to the serial adapter.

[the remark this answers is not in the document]
jwright6323: IMO `SERIAL_IF_WIDTH` should be `XLen` but I think that change is outside the scope of this PR. I'm not opposed to leaving `mmioRegWidth` as long as there's an assertion. Also yes, `SERIAL_IF_WIDTH` is named poorly, but that's not your fault :). I think `serialIfWidth` is the appropriate name for what it does.

src/main/scala/tsiHost/TsiHostWidget.scala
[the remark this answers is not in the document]
jwright6323: See below, I think we should eventually make this all `XLen` but for now passing it to the `RegField` seems right.

abejgonzalez:
Added a bit more comments and checked that this works with `master` of `project-template`. Note: this also works with the `eagle` repo too.

abejgonzalez:
Close since #113 cleans up this PR

PR #35, opened 2019-03-02
Don't do anything special with simserial args

### LINKED ISSUE #34: SimSerial doesn't handle +permissive

colinschmidt:
Fesvr handles this now so just pass on the plusargs and
they will be correctly ignored
Fixes #34


colinschmidt:
Can I merge this? @grebe @zhemao

grebe:
I think this is fine

PR #36, opened 2019-03-02
Fix periodchecker to not use timescale

colinschmidt:
The timescale directive isn't a local change so the propsed method avoids changing other bits of code which might rely on the default timescale.


colinschmidt:
I assume no one cares about this so I'm going to merge it at end of day. @jwright6323

PR #38, opened 2019-03-07
Add prefix to vsrc, csrc and add bootrom to resources

grebe:


grebe:
I don't know how I missed this before... project-template was relying on this and I just forgot to push it

PR #42, opened 2019-04-10
Update Switcher for rocket-chip Feb 14 2019

colinschmidt:
Single mBus multi-port switcher support

rocket-chip now uses a single mBus with multiple nested ports for multiple memory channels so the canonical way of connecting the switcher would be with a `:*=` connection.
Without this change this will result in overlapping address spaces and fail in diplomacy.

### DIFF
--- a/src/main/scala/Switcher.scala
+++ b/src/main/scala/Switcher.scala
@@ -103,10 +103,10 @@ class TLSwitcher(
 
   val device = new SimpleDevice("switcher", Seq("ucb-bar,switcher"))
 
-  val innode = TLManagerNode(Seq.fill(inPortN) {
+  val innode = TLManagerNode(Seq.tabulate(inPortN) { i =>
     TLManagerPortParameters(
       Seq(TLManagerParameters(
-        address    = address,
+        address    = Seq(address(i)),
         resources  = device.reg("mem"),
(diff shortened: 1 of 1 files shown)


src/main/scala/Switcher.scala
```diff
@@ -103,10 +103,10 @@ class TLSwitcher(
 
   val device = new SimpleDevice("switcher", Seq("ucb-bar,switcher"))
 
-  val innode = TLManagerNode(Seq.fill(inPortN) {
+  val innode = TLManagerNode(Seq.tabulate(inPortN) { i =>
     TLManagerPortParameters(
       Seq(TLManagerParameters(
-        address    = address,
+        address    = Seq(address(i)),
```
jwright6323: This is the only place where it looks like behavior might have changed. Now each `TLManagerParameters` object is getting a 1-long `Seq` of `AddressSet` rather than a `Seq` of all `AddressSets`. The new style seems more correct- how did it work before? Is it because we always gave it a 1-long `AddressSet`?

colinschmidt: We used to have an mbus per memory channel and do a loop through them such that each connection was essentially for a 1-long `Seq[AddressSet]`

PR #45, opened 2019-05-23
SERDES Fix

abejgonzalez:
Have the SERDES take the max params of the client and the manager.


abejgonzalez:
@jwright6323 @colinschmidt Just a reminder after the tape out is done to take a quick look at this!

abejgonzalez:
Is this good to merge in? @colinschmidt @jwright6323

jwright6323:
@abejgonzalez I think this is actually broken. We needed to make some changes to get it to work on EAGLE. @colinschmidt has it on a branch, I believe.

colinschmidt:
We didn't make any changes to this for my tapeout.
I believe there is some code change that would clear up edge cases but want to double check that(in a bunch of corner cases) and will make it a separate PR.

PR #46, opened 2019-06-19
update for debug apb & regiontype in rocket-chip

harrisonliew:
https://github.com/freechipsproject/rocket-chip/commit/4c5b38edbcd6eb12a342982c1fcd99ece550037d made dmi optional, and https://github.com/chipsalliance/rocket-chip/pull/1940 split the region type.
Needed for recent rocket-chip versions but will break with older ones.

### DIFF
--- a/src/main/scala/Switcher.scala
+++ b/src/main/scala/Switcher.scala
@@ -109,7 +109,7 @@ class TLSwitcher(
         address    = address(i),
         resources  = device.reg("mem"),
         regionType = if (cacheable) RegionType.UNCACHED
-                     else RegionType.UNCACHEABLE,
+                     else RegionType.IDEMPOTENT,
         executable = executable,
         supportsGet        = TransferSizes(1, lineBytes),
         supportsPutPartial = TransferSizes(1, lineBytes),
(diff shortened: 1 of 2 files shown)


src/main/scala/Switcher.scala
```diff
@@ -109,7 +109,7 @@ class TLSwitcher(
         address    = address(i),
         resources  = device.reg("mem"),
         regionType = if (cacheable) RegionType.UNCACHED
-                     else RegionType.UNCACHEABLE,
+                     else RegionType.IDEMPOTENT,
```
jwright6323: We might want this to be `VOLATILE`

jwright6323:
This LGTM but I'm going to let @colinschmidt review before I hit the button. IDK what the use case for the switcher on uncached memory would be, so I'm not sure if it should be `IDEMPOTENT` or `VOLATILE`. It probably doesn't matter.

colinschmidt:
Well someone is going to have to live on a branch. I think it makes most sense for me to live on the branch since I'm on the old rocket-chip.

src/main/scala/Switcher.scala
colinschmidt: Yeah I feel like the common use for this would be some offchip memory that could be anything so setting it to idempotent seems less safe.

abejgonzalez:
Is this good to merge in? @colinschmidt @jwright6323

jwright6323:
@abejgonzalez @harrisonliew No. At a minimum `IDEMPOTENT` needs to be `VOLATILE`

abejgonzalez:
Other than the `VOLATILE` change... what else needs to be done to fix this PR?

PR #47, opened 2019-06-26
comments on bootrom

abejgonzalez:


src/main/resources/testchipip/bootrom/bootrom.S
```diff
@@ -16,33 +18,36 @@ interrupt_loop:
   bnez a3, interrupt_loop
   j boot_core_hart0
 
+// enter here at first and wait for MSIP interrupt to start program execution
```
a0u: I suggest moving the "wait for MSIP interrupt" part to line 33, where the actual loop is.  Also, a more precise term for describing `_hang` is "reset vector".

PR #51, opened 2019-08-01
Add "override lazy" to Generator.longName to sync with rocket-chip update.

ucbjrl:
See chipsalliance/rocket-chip#2049

### DIFF
--- a/src/main/scala/Generator.scala
+++ b/src/main/scala/Generator.scala
@@ -7,7 +7,7 @@ import java.io.{File, FileWriter}
 import firrtl.annotations.JsonProtocol
 
 object Generator extends GeneratorApp {
-  val longName = names.topModuleProject + "." + names.topModuleClass + "." + names.configs
+  override lazy val longName = names.topModuleProject + "." + names.topModuleClass + "." + names.configs
   generateFirrtl
   generateAnno
 }


abejgonzalez:
Should this change be echoed in the Chipyard area (generators/example/...)? Also, has this been confirmed to work with Chipyard? Finally, what is this Generator used for?

ucbjrl:
Great questions @abejgonzalez. @grebe added this repository to [chisel-release](https://github.com/ucb-bar/chisel-release) (ucb-bar/chisel-release#5) on the assumption that it would be published with rocket-chip. If it has more dependencies, it should probably be removed from chisel-release, and managed along with its dependencies. (I'm still not sure chisel-release should include rocket-chip. Although we're currently publishing master SNAPSHOTS, I think this role should be moved to Chipyard or SiFive).

abejgonzalez:
The only dependency is Rocket Chip so it should be fine to bundle with `chisel-release`. I still don't follow why there is a generator here? My impression is that the SoC projects (`rocketchip`, `chipyard`, etc) will have their own generators which will be invoked. There will not be a Generator invoked from `testchipip`.

ucbjrl:
Presumably handled elsewhere.

PR #52, opened 2019-08-07
Fixes following removal of old chisel3 deprecations.

ucbjrl:
Replace Vecf(init) with VecInit()
Replace Wire(init) with WireDefault()
Replace Module(clock, reset) with withClockAndReset
Replace chiselCloneType with chiselTypeOf
Add asUInt to expressions that now return Bits

Don't merge this commit until we verify that the chisel3 changes are desirable.

### DIFF
--- a/src/main/scala/Util.scala
+++ b/src/main/scala/Util.scala
@@ -60,19 +60,15 @@ case class AsyncWideCounter(width: Int, inc: UInt = 1.U, reset: Boolean = true)
 }
 
 // As WideCounter, but it's a module so it can take arbitrary clocks
-class WideCounterModule(w: Int, inc: UInt = 1.U, reset: Boolean = true, clockSignal: Clock = null, resetSignal: Bool = null)
+class WideCounterModule(w: Int, inc: UInt = 1.U, doReset: Boolean = true, clockSignal: Clock = null, resetSignal: Bool = null)
     extends Module {
   val io = IO(new Bundle {
     val value = Output(UInt(w.W))
   })
-  lazy val block = {
-    io.value := AsyncWideCounter(w, inc, reset).value
-  }
-  (clockSignal, resetSignal) match {
-    case (null, null) => block
-    case (null, r: Bool) => withReset(r) { block }
-    case (c: Clock, null) => withClock(c) { block }
-    case (c: Clock, r: Bool) => withClockAndReset(c, r) { block }
+  io.value := {
+    val cc = Option(clockSignal).getOrElse(this.clock)
+    val rr = Option(resetSignal).getOrElse(this.reset)
+    withClockAndReset(cc, rr) { AsyncWideCounter(w, inc, doReset).value }
   }
 }
 
@@ -114,7 +110,7 @@ class BinToGray[T <: Data](gen: T, c: Clock) extends Module {
     val gray = UInt(gen.getWidth.W)
   })
   withClock(c) {
-    io.gray := RegNext(io.bin.asUInt ^ (io.bin.asUInt >> 1.U).asUInt())
+    io.gray := RegNext(io.bin.asUInt ^ (io.bin.asUInt >> 1.U).asUInt)
   }
 }
 
--- a/src/main/scala/Util.scala
+++ b/src/main/scala/Util.scala
@@ -7,12 +7,14 @@ import freechips.rocketchip.diplomacy.{IdRange, ValName}
 import freechips.rocketchip.util.AsyncResetReg
 import freechips.rocketchip.tilelink._
 
-class ResetSync(c: Clock, lat: Int = 2) extends Module(_clock = c) {
+class ResetSync(c: Clock, lat: Int = 2) extends Module {
   val io = IO(new Bundle {
     val reset = Input(Bool())
     val reset_sync = Output(Bool())
   })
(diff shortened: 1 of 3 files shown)


src/main/scala/Util.scala
```diff
@@ -84,25 +94,28 @@ object WideCounterModule {
...
   bin2gray.io.bin := io.in
   val out_gray = ShiftRegister(bin2gray.io.gray, lat)
-  io.out := gen.cloneType.fromBits(
+  io.out := (
     (0 until size)
-      .map(out_gray.asUInt >> _.U)
-      .reduceLeft((a: UInt, b: UInt) => a^b))
+      .map{ case og => (out_gray.asUInt >> og.U).asUInt }
+      .reduceLeft((a: UInt, b: UInt) => (a^b).asUInt)
+    )
```
seldridge: This looks fine, but I'm only inspecting. I haven't tested any of this...

src/main/scala/Util.scala
```diff
@@ -84,25 +94,28 @@ object WideCounterModule {
...
-class BinToGray[T <: Data](gen: T, c: Clock) extends Module(_clock = c) {
+class BinToGray[T <: Data](gen: T, c: Clock) extends Module {
   val io = IO(new Bundle {
-    val bin = Flipped(gen.chiselCloneType)
+    val bin = Flipped(chiselTypeOf(gen))
     val gray = UInt(gen.getWidth.W)
   })
-  io.gray := Reg(next=(io.bin.asUInt ^ (io.bin.asUInt >> 1.U)))
+  withClock(c) {
+    io.gray := RegNext(io.bin.asUInt ^ (io.bin.asUInt >> 1.U).asUInt())
```
seldridge:
```suggestion
    io.gray := RegNext(io.bin.asUInt ^ (io.bin.asUInt >> 1.U).asUInt)
```

src/main/scala/Util.scala
```diff
@@ -59,11 +61,19 @@ case class AsyncWideCounter(width: Int, inc: UInt = 1.U, reset: Boolean = true)
...
-  io.value := AsyncWideCounter(w, inc, reset).value
+  lazy val block = {
+    io.value := AsyncWideCounter(w, inc, reset).value
+  }
+  (clockSignal, resetSignal) match {
+    case (null, null) => block
+    case (null, r: Bool) => withReset(r) { block }
+    case (c: Clock, null) => withClock(c) { block }
+    case (c: Clock, r: Bool) => withClockAndReset(c, r) { block }
+  }
```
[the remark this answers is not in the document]
ucbjrl: Much better. But there's potential confusion with `reset` since it's a parameter to this class. How about:
```scala
  io.value := {
    val cc = Option(clockSignal).getOrElse(super.clock)
    val rr = Option(resetSignal).getOrElse(super.reset)
    withClockAndReset(cc, rr) { AsyncWideCounter(w, inc, reset).value }
  }
```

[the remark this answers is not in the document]
ucbjrl: Nope.
```
Error:(69, 50) super may not be used on value clock
    val cc = Option(clockSignal).getOrElse(super.clock)
Error:(70, 50) super may not be used on value reset
    val rr = Option(resetSignal).getOrElse(super.reset)
```

src/main/scala/Util.scala
```diff
@@ -7,12 +7,14 @@ import freechips.rocketchip.diplomacy.{IdRange, ValName}
...
+class ResetSync(c: Clock, lat: Int = 2) extends Module {
   val io = IO(new Bundle {
     val reset = Input(Bool())
     val reset_sync = Output(Bool())
   })
-  io.reset_sync := ShiftRegister(io.reset,lat)
+  withClock(c) {
+    io.reset_sync := ShiftRegister(io.reset,lat)
+  }
 }
```
jwright6323: @colinschmidt @zhemao Do we even use this `ResetSync`? It seems redundant with things that exist in rocket-chip, which we import anyway.

src/main/scala/Util.scala
jwright6323: I think this is beyond the scope of this PR, but this entire class looks broken to me. Gray coding is only safe for clock crossings if you're incrementing by 1, which does not appear to be what this is doing...

src/main/scala/Util.scala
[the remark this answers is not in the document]
ucbjrl: The _obvious_ solution:
```scala
class WideCounterModule(w: Int, inc: UInt = 1.U, doReset: Boolean = true, clockSignal: Clock = null, resetSignal: Bool = null)
    extends Module {
  val io = IO(new Bundle {
    val value = Output(UInt(w.W))
  })
  io.value := {
    val cc = Option(clockSignal).getOrElse(this.clock)
    val rr = Option(resetSignal).getOrElse(this.reset)
    withClockAndReset(cc, rr) { AsyncWideCounter(w, inc, doReset).value }
  }
}
```

src/main/scala/Util.scala
colinschmidt: I always use `ResetCatchAndSync`, assuming these two do the same thing.

ucbjrl:
Presumably handled elsewhere.

PR #53, opened 2019-08-14
Bump sbt to something less dangerous and don't ignore non-target project files.

ucbjrl:


colinschmidt:
Do all consuming projects need to bump sbt at the same time?

ucbjrl:
I don't believe it's strictly necessary, although to minimize the effect of potential sbt weirdness, we should try to keep them consistent.

Ideally, this can be packaged as a library and the version of sbt used to build it should have no impact on its use by clients.

Perhaps we should discuss packaging and publishing before this PR is approved (or rejected).

kammoh:
Any reasons not upgrading to 1.3.0?

ucbjrl:
I'd like to see firrtl and chisel3 upgraded to sbt 1.3.0 first.

ucbjrl:
Presumably handled elsewhere.

PR #54, opened 2019-08-20
Exposure updates

ucbjrl:
This depends on freechipsproject/chisel3#1162


colinschmidt:
We should make sure this works with chipyard before we merge this.
It might be too early now but we should eventually make a PR that is based on chipyard/dev and bumps the testchipip pointer there to get CI to run.

ucbjrl:
Presumably handled elsewhere.

PR #55, opened 2019-08-21
Revert "remove header from resources"

zhemao:
This reverts commit 53c019a4925cef967a3b1cc6a51c7d74a9aa677f.


zhemao:
Wait, ack, now you've messed up the submodule pointer in chipyard.

jerryz123:
We should disable rebase and squash commits on all chipyard submodules

jwright6323:
Those should be fine as long as you point chipyard to the merged commit. I don't think you want to point to a non-dev/master branch in chipyard anyway.

PR #59, opened 2019-11-25
Add a UART Adapter

abejgonzalez:
Adds a new trait that can be used in projects to create a UART and a corresponding UARTAdapter on the TestHarness side. By adding the plusarg `uartlog` you can create an output file(s) that gets the redirected `stdout`. I haven't tested in the `stdin` feature of this adapter but that was out of scope for this.

Todo: 
- [x] Have the output go to a file instead of `stdout`

### DIFF
--- a/src/main/resources/testchipip/csrc/SimUART.cc
+++ b/src/main/resources/testchipip/csrc/SimUART.cc
@@ -6,7 +6,7 @@
 
 #include "uart.h"
 
-uart_t *uart = 0;
+uart_t *uart = NULL;
 
 extern "C" void uart_init(
         const char *filename,
@@ -27,7 +27,7 @@ extern "C" void uart_tick(
         unsigned char in_ready,
         char *in_bits)
 {
-    if (uart == 0) {
+    if (uart == NULL) {
         *out_ready = 0;
         *in_valid = 0;
         return;
--- a/src/main/resources/testchipip/csrc/SimUART.cc
+++ b/src/main/resources/testchipip/csrc/SimUART.cc
@@ -1,0 +1,39 @@
+#include <vpi_user.h>
+#include <svdpi.h>
+
+#include "uart.h"
+#include <stdio.h>
+
+uart_t *uart = 0;
+
+extern "C" void uart_init(
+        const char *filename,
+        int uartno)
+{
+    uart = new uart_t(filename, uartno);
+}
+
+extern "C" void uart_tick(
+        unsigned char out_valid,
+        unsigned char *out_ready,
+        char out_bits,
+
+        unsigned char *in_valid,
+        unsigned char in_ready,
+        char *in_bits)
+{
+    if (uart == 0) {
+        *out_ready = 0;
+        *in_valid = 0;
+        return;
+    }
+
+    uart->tick(
+        out_valid,
+        out_ready,
+        out_bits,
+
+        in_valid,
+        in_ready,
+        in_bits);
+}
--- a/src/main/resources/testchipip/csrc/SimUART.cc
+++ b/src/main/resources/testchipip/csrc/SimUART.cc
@@ -1,8 +1,10 @@
 #include <vpi_user.h>
 #include <svdpi.h>
 
-#include "uart.h"
 #include <stdio.h>
+#include <string.h>
+
+#include "uart.h"
 
 uart_t *uart = 0;
 
@@ -10,7 +12,10 @@ extern "C" void uart_init(
         const char *filename,
         int uartno)
 {
-    uart = new uart_t(filename, uartno);
+    if (strlen(filename) != 0)
+        uart = new uart_t(filename, uartno);
+    else
+        uart = new uart_t(0, uartno);
 }
 
 extern "C" void uart_tick(
--- a/src/main/resources/testchipip/csrc/uart.cc
+++ b/src/main/resources/testchipip/csrc/uart.cc
@@ -33,6 +33,8 @@ void sighand(int s) {
             specialchar = 0x0;
     }
 }
+#else
+#warning "The UARTAdapter is untested with Windows. Many features expected to be broken."
 #endif
 
 uart_t::uart_t(const char* filename_prefix, int uartno)
@@ -58,22 +60,22 @@ uart_t::uart_t(const char* filename_prefix, int uartno)
         this->outputfd = STDOUT_FILENO;
     } else {
         // for UARTs that are not UART0, use a PTY
-        char slavename[PTYNAMELEN];
+        char ptyname[PTYNAMELEN];
         int ptyfd = posix_openpt(O_RDWR | O_NOCTTY);
         grantpt(ptyfd);
         unlockpt(ptyfd);
-        ptsname_r(ptyfd, slavename, PTYNAMELEN);
+        ptsname_r(ptyfd, ptyname, PTYNAMELEN);
 
         // create symlink for reliable location to find uart pty
         std::string symlinkname = std::string("uartpty") + std::to_string(uartno);
-        // unlink in case symlink already exists
-        unlink(symlinkname.c_str());
-        if(symlink(slavename, symlinkname.c_str())) {
-            printf("[UART_ERR] Failed to created symlink with slave %s to symlink name %s\n", slavename, symlinkname.c_str());
+        // remove in case symlink already exists
+        remove(symlinkname.c_str());
+        if(symlink(ptyname, symlinkname.c_str())) {
+            printf("[UART_ERR] Failed to created symlink with slave %s to symlink name %s\n", ptyname, symlinkname.c_str());
             exit(1);
         }
-        printf("[UART] UART%d is on PTY: %s, symlinked at %s\n", uartno, slavename, symlinkname.c_str());
-        printf("[UART] Attach to this UART with 'sudo screen %s' or 'sudo screen %s'\n", slavename, symlinkname.c_str());
+        printf("[UART] UART%d is on PTY: %s, symlinked at %s\n", uartno, ptyname, symlinkname.c_str());
+        printf("[UART] Attach to this UART with 'sudo screen %s' or 'sudo screen %s'\n", ptyname, symlinkname.c_str());
         this->inputfd = ptyfd;
         this->outputfd = ptyfd;
     }
--- a/src/main/resources/testchipip/csrc/uart.cc
+++ b/src/main/resources/testchipip/csrc/uart.cc
@@ -12,7 +12,7 @@
 #include <unistd.h>
 
 // name length limit for ptys
-#define SLAVENAMELEN 256
+#define PTYNAMELEN 256
 
 /* There is no "backpressure" to the user input for sigs. only one at a time
  * non-zero value represents unconsumed special char input.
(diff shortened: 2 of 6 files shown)


src/main/scala/UARTAdapter.scala
```diff
@@ -0,0 +1,149 @@
...
+    val serial = Flipped(new SerialIO(width))
+  })
+
+  addResource("/testchipip/vsrc/SimUART.v")
+  addResource("/testchipip/csrc/SimUART.cc")
+  addResource("/testchipip/csrc/uart.cc")
+  addResource("/testchipip/csrc/uart.h")
+}
+
+trait HasPeripheryUARTWithAdapter extends HasPeripheryUART { this: BaseSubsystem =>
```
jerryz123: This should be named `CanHavePeripheryUartAdapter`

src/main/scala/UARTAdapter.scala
```diff
@@ -0,0 +1,149 @@
...
+}
+
+trait HasPeripheryUARTWithAdapterImp extends HasPeripheryUARTModuleImp {
```
jerryz123: This should also be `CanHavePeripheryUARTAdapterModuleImp`

abejgonzalez:
I think the main problem with this is that this will make `testchipip` depend on `sifive-blocks`. So if you don't build within a top level setting (i.e. Chipyard) this will break if it can't see `sifive-blocks`.

jerryz123:
> I think the main problem with this is that this will make `testchipip` depend on `sifive-blocks`. So if you don't build within a top level setting (i.e. Chipyard) this will break if it can't see `sifive-blocks`.

You can get around this by replacing `package testchipip` with `package sifive.blocks.devices.uart` at the top of the file.

src/main/scala/UARTAdapter.scala
abejgonzalez: Ok changed them to `CanHave`s. Note, the name of the trait is "have UART with Adapter". not "have UART adapter" to note that this adds both the UART and the corresponding adapter. The empty trait is to add the main `HasPeripheryUART` trait while having similar naming to the `Imp` trait.

src/main/scala/UARTAdapter.scala
abejgonzalez: Done. See above.

abejgonzalez:
> > I think the main problem with this is that this will make `testchipip` depend on `sifive-blocks`. So if you don't build within a top level setting (i.e. Chipyard) this will break if it can't see `sifive-blocks`.
> 
> You can get around this by replacing `package testchipip` with `package sifive.blocks.devices.uart` at the top of the file.

While that might be fine... I think we should also know who uses `testchipip` by pulling from Maven.

src/main/resources/testchipip/csrc/uart.cc
```diff
@@ -0,0 +1,138 @@
...
+
+#define _XOPEN_SOURCE
+#include <stdlib.h>
+#include <stdio.h>
+
+#ifndef _WIN32
+#include <unistd.h>
+
+// name length limit for ptys
+#define SLAVENAMELEN 256
```
colinschmidt: Can we use "PTYNAMELEN" instead?

abejgonzalez:
> How did you test this?

I tested this with Linux boot (`br-base-bin`) and with a sample UART program (based on the UART/TTY headers in Beagle/Eagle bootrom) that output a bunch of sentences to the screen. 

The Linux boot was successful (ran in cosimulation) but I think the baudrate was off (I was seeing multiple characters outputted to the screen). I was going to retry this later.

The sample UART program worked nicely. I was also able to redirect its output to a file using the `+uartlog`.

Note: I haven't tested `stdin` just `stdout`

abejgonzalez:
> The Linux boot was successful (ran in cosimulation) but I think the baudrate was off (I was seeing multiple characters outputted to the screen). I was going to retry this later.

As an update, the baudrate was fine. The discrepancy was the co-simulator that I would using in conjunction with the UART Adapter.

abejgonzalez:
@colinschmidt @jerryz123 Does this look fine? Do you all want me to switch the package to `package sifive.blocks.devices.uart` for this block since it depends on having a UART?

jerryz123:
This looks fine to me, but other people should confirm

abejgonzalez:
Ok I changed the package to `package sifive.blocks.devices.uart` so that this adapter doesn't build unless `sifive-blocks` exists.

src/main/resources/testchipip/csrc/SimUART.cc
```diff
@@ -0,0 +1,44 @@
...
+extern "C" void uart_tick(
+        unsigned char out_valid,
+        unsigned char *out_ready,
+        char out_bits,
+
+        unsigned char *in_valid,
+        unsigned char in_ready,
+        char *in_bits)
+{
+    if (uart == 0) {
```
colinschmidt: Might be worth making this guard consistent with the other guards in SimSerial and SimBlockDevice.
Initialize at the top to `NULL` and then here check `!uart` or `uart == NULL` (serial vs blockdevice).

src/main/resources/testchipip/csrc/uart.cc
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * Reset to zero once consumed.
+ */
+
+// This is fine for multiple UARTs because UARTs > uart 0 will use pty, not stdio
+char specialchar = 0;
+
+void sighand(int s) {
+    switch (s) {
+        case SIGINT:
```
colinschmidt: Aren't people going to be annoyed when other signals are ignored?
Or is there some reason we can ignore other signals (like ctrl-D)?

src/main/resources/testchipip/csrc/uart.cc
```diff
@@ -0,0 +1,138 @@
...
+            // ctrl-c
+            specialchar = 0x3;
+            break;
+        default:
+            specialchar = 0x0;
+    }
+}
+#endif
```
colinschmidt: Should we offer some (compile-time)warning that some features are not supported on WIN32?

src/main/resources/testchipip/csrc/uart.cc
```diff
@@ -0,0 +1,138 @@
...
+        sigIntHandler.sa_handler = sighand;
+        sigemptyset(&sigIntHandler.sa_mask);
+        sigIntHandler.sa_flags = 0;
+        sigaction(SIGINT, &sigIntHandler, NULL);
+        printf("[UART] UART0 is here (stdin/stdout).\n");
+        this->inputfd = STDIN_FILENO;
+        this->outputfd = STDOUT_FILENO;
+    } else {
+        // for UARTs that are not UART0, use a PTY
+        char slavename[PTYNAMELEN];
```
colinschmidt: Lets just call this ptyname.

src/main/resources/testchipip/csrc/uart.cc
```diff
@@ -0,0 +1,138 @@
...
+        int ptyfd = posix_openpt(O_RDWR | O_NOCTTY);
+        grantpt(ptyfd);
+        unlockpt(ptyfd);
+        ptsname_r(ptyfd, slavename, PTYNAMELEN);
+
+        // create symlink for reliable location to find uart pty
+        std::string symlinkname = std::string("uartpty") + std::to_string(uartno);
+        // unlink in case symlink already exists
+        unlink(symlinkname.c_str());
```
colinschmidt: It looks like `remove` is more portable than `unlink`, but I don't really care.

src/main/resources/testchipip/csrc/uart.cc
```diff
@@ -0,0 +1,138 @@
...
+        if(symlink(slavename, symlinkname.c_str())) {
+            printf("[UART_ERR] Failed to created symlink with slave %s to symlink name %s\n", slavename, symlinkname.c_str());
+            exit(1);
+        }
+        printf("[UART] UART%d is on PTY: %s, symlinked at %s\n", uartno, slavename, symlinkname.c_str());
+        printf("[UART] Attach to this UART with 'sudo screen %s' or 'sudo screen %s'\n", slavename, symlinkname.c_str());
```
colinschmidt: Is `sudo` necessary here?

src/main/scala/UARTAdapter.scala
```diff
@@ -0,0 +1,156 @@
...
+}
+import UARTAdapterConsts._
+
+case object UARTAdapterKey extends Field[Seq[UARTAdapterParams]](Nil)
+
+case class UARTAdapterParams(
+  baudRateInit: BigInt = BigInt(115200)
+)
+
+class UARTAdapter(uartno: Int)(implicit p: Parameters) extends Module
```
colinschmidt: Why do we want this adapter rather than the normal sifive-block uart?
Does the sim part also work with that uart?
Assuming we do want this adapter is there a way to reduce the code duplication?

abejgonzalez: Sorry this may not be clear, and Ill add some comments to this. This adapter connects to each SiFive UART in the TestHarness (1 adapter to 1 sifive uart). So for this to work, each configuration needs to have at least 1 SiFive UART.

As for code duplication, this is adapted from the `UARTBridge` made for FireSim... so there was already code duplication there. I think that can be combined/deduped a bit, however, I don't see an obvious way to reduce duplication with the SiFive UART.

src/main/resources/testchipip/csrc/SimUART.cc
abejgonzalez: Sure. Thats fine.

src/main/resources/testchipip/csrc/uart.cc
abejgonzalez: In the Chipyard documentation (where I assume most people will use this), we state that Windows is not recommended... so that should deter most people from using Windows. However, we can add a warning here for Windows. Something like `WARNING: The UARTAdapter instantiated is untested with Windows. Many features expected to be broken`

src/main/resources/testchipip/csrc/uart.cc
abejgonzalez: I honestly haven't tried this particular feature since I haven't made it to the prompt of Linux boot.

src/main/scala/UARTAdapter.scala
```diff
@@ -0,0 +1,156 @@
...
+  addResource("/testchipip/vsrc/SimUART.v")
+  addResource("/testchipip/csrc/SimUART.cc")
+  addResource("/testchipip/csrc/uart.cc")
+  addResource("/testchipip/csrc/uart.h")
+}
+
+trait CanHavePeripheryUARTWithAdapter extends HasPeripheryUART { this: BaseSubsystem =>
+}
+
+trait CanHavePeripheryUARTWithAdapterImp extends HasPeripheryUARTModuleImp {
```
jerryz123: One way to avoid having this extra trait (and the confusion between this trait and the base `HasPeripheryUART` trait) is to make the `connectSimUARTs` function a global function, as opposed to a method of the `CanHavePeripheryUARTWithAdapterImp`. The function could be 

```
object ConnectUARTPrinter {
  def apply[T <: HasPeripheryUARTModuleImp](top: T, p: Parameters): Unit = {
    require(p(PeripheryUARTKey).size == p(UARTAdapterKey).size)
    top.uart.zipWithIndex.foreach{ case (dut_io, i) =>
      val uart_sim = Module(new UARTAdapter(i)(p))
      uart_sim.io.uart.txd := dut_io.txd
      dut_io.rxd := uart_sim.io.uart.rxd
  }
}
```

You can then call this function in Chipyard's testharness on the dut returned by `BuildTop`.
This seems to me to be a bit more elegant than adding the additional trait, since traits typically describe additional hardware that is added to the system, but I may be missing some sublety.

abejgonzalez: I'm not opposed to either approach. The main reason I had this as a trait was to make it similar to the block device and the serial adapter (a single trait that adds both the DUT component as well as the TestHarness component).

src/main/resources/testchipip/csrc/uart.cc
abejgonzalez: Again, the purpose of this widget is mainly to read from the UART and not interact with it. With that being said, this is currently what FireSim does so I copied that. @sagark might know a particular reason why this is ignoring most of the other signals.

src/main/scala/UARTAdapter.scala
abejgonzalez: Now that I think about this more, I think I prefer having the current approach, a separate trait from the UART trait that can potentially add the UARTAdapters in.

PR #61, opened 2019-12-10
Chisel update aug 2019

colinschmidt:
Updates for rocket-chip and chisel 3.2.0 for chipyard see: ucb-bar/chipyard#358


colinschmidt:
Self approving and merging.

PR #66, opened 2020-02-25
Add SimDRAM BlackBox

zhemao:
This makes simulating larger memories in chipyard more feasible than with the SimAXIMem RTL model.


src/main/scala/SimDRAM.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  require(params.dataBits <= 64)
+
+  addResource("/testchipip/vsrc/SimDRAM.v")
+  addResource("/testchipip/csrc/SimDRAM.cc")
+  addResource("/testchipip/csrc/mm.cc")
+  addResource("/testchipip/csrc/mm.h")
+}
+
+trait CanHaveBlackBoxSimMemModuleImp { this: CanHaveMasterAXI4MemPortModuleImp =>
```
jerryz123: Is it necessary for this to be a trait? Why can't we collapse this entirely within an IOBinder?
```
class WithBlackBoxSimMem extends OverrideIOBinder({
  (c, r, s, top: CanHaveMasterAXI4MemPortModuleImp) => {
    val p = top.p
    (top.mem_axi4 zip top.outer.memAXI4Node).foreach { case (io, node) =>
      val memSize = p(ExtMem).get.master.size
      val lineSize = p(CacheBlockBytes)
      (io zip node.in).foreach { case (axi4, (_, edge)) =>
        val mem = Module(new SimDRAM(memSize, lineSize, edge.bundle))
        mem.io.axi <> axi4
        mem.io.clock := clock
        mem.io.reset := reset
      }
    }
  }
})
```

zhemao: Oh, I guess that's true.

a0u:
SimDRAM is a bit of a misnomer if it is not actually model DRAM timings.

Instead of re-implementing the code, can this be unified with the MIDAS wrappers for magic memory and DRAMSim?

a0u:
Proof of concept: https://github.com/ucb-bar/chipyard/tree/cs152-sp20-dramsim2

src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -0,0 +1,150 @@
...
+          NULL, sz, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, -1, 0);
+  size = sz;
+}
+
+mm_t::~mm_t()
+{
+  munmap(data, this->size);
+}
+
+void mm_magic_t::init(size_t sz, int wsz, int lsz)
```
colinschmidt: Why do we need to reimplement a magic AXI memory?

src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -0,0 +1,150 @@
...
+  cycle++;
+
+  if (reset) {
+    while (!bresp.empty()) bresp.pop();
+    while (!rresp.empty()) rresp.pop();
+    cycle = 0;
+  }
+}
+
+void load_mem(void** mems, const char* fn, int line_size, int nchannels)
```
colinschmidt: Is this the same format as `readmemh`?

src/main/scala/SimDRAM.scala
```diff
@@ -0,0 +1,33 @@
...
+  val io = IO(new Bundle {
+    val clock = Input(Clock())
+    val reset = Input(Bool())
+    val axi = Flipped(new AXI4Bundle(params))
+  })
+
+  require(params.dataBits <= 64)
+
+  addResource("/testchipip/vsrc/SimDRAM.v")
+  addResource("/testchipip/csrc/SimDRAM.cc")
```
colinschmidt: Was this replicated in chipyard?

src/main/resources/testchipip/csrc/mm.cc
zhemao: Per chipyard meeting discussion. Simulating magic memory in a C++ model is faster than simulating SimAXIMem RTL.

src/main/scala/SimDRAM.scala
zhemao: What do you mean replicated?

src/main/resources/testchipip/csrc/mm.cc
zhemao: I believe so. But no spaces or comments. I just copied this from the old rocketchip code.

PR #67, opened 2020-02-28
TileLink Ring  Network

zhemao:
This implements a ring topology network for TileLink and a SystemBus subclass based on it. This allows chipyard/firesim designs to easily sub in the ring for the default crossbar system bus. The topology-independent code is factored out so that additional topologies can be implemented later on.


src/main/scala/Ring.scala
```diff
@@ -0,0 +1,185 @@
...
+      }
+    } else {
+      io_out.head <> io_in.head
+    }
+  }
+}
+
+class RingSystemBus(params: SystemBusParams, buffer: TLNetworkBufferParams)
+    (implicit p: Parameters) extends SystemBus(params) {
+  private val system_bus_ring = LazyModule(new TLRingNetwork(buffer))
```
jwright6323: If this is private can Chisel still name it sanely?

zhemao: Yes it does. I checked the generated verilog and it has the right name.

jwright6323: I guess I should have asked this before, but is there a particular reason it needs to be private? Correct me if I'm wrong, but this would prevent us from accessing this via an aspect wouldn't it? I could see us wanting to do that with the floorplanning framework.

src/main/scala/Ring.scala
```diff
@@ -0,0 +1,207 @@
...
+  val nNodes = nIn + nOut
+  val io = IO(new NetworkIO(nIn, nOut, payloadTyp, Some(nNodes)))
+
+  def mappingOK(mapping: Seq[Int], n: Int): Boolean = {
+    mapping.size == n &&
+    mapping.sorted == Seq.tabulate(n)(i => i)
+  }
+
+  require(inputMap.isEmpty  || mappingOK(inputMap.get, nIn))
+  require(outputMap.isEmpty || mappingOK(outputMap.get, nOut))
```
jwright6323: It would be nice to tell the user why the `require` fails. Something like:
```suggestion
  require(outputMap.isEmpty || mappingOK(outputMap.get, nOut), "outputMap must either be None or the same length as the number of outputs")
```

src/main/scala/Ring.scala
```diff
@@ -0,0 +1,207 @@
...
+    extends NetworkInterconnect[T] {
+  val nNodes = nIn + nOut
+  val io = IO(new NetworkIO(nIn, nOut, payloadTyp, Some(nNodes)))
+
+  def mappingOK(mapping: Seq[Int], n: Int): Boolean = {
+    mapping.size == n &&
+    mapping.sorted == Seq.tabulate(n)(i => i)
+  }
+
+  require(inputMap.isEmpty  || mappingOK(inputMap.get, nIn))
```
jwright6323: Same comment:
```suggestion
  require(inputMap.isEmpty  || mappingOK(inputMap.get, nIn), "inputMap must either be None or the same length as the number of inputs")
```

src/main/scala/Ring.scala
zhemao: @jwright6323 I pushed a commit to add a more descriptive message. Do you think it's satisfactory?

jwright6323: LGTM

src/main/scala/Ring.scala
zhemao: The equivalent structure in the standard SystemBus was private, and I thought there was some reason for that, but apparently not.

PR #69, opened 2020-03-14
Modularize connect methods

jwright6323:
This takes the connect methods from SerialAdapter, BlockDevice, and UARTAdapter and puts them into objects so that they can be easily reused by Top and ChipTop (See ucb-bar/chipyard#480)

### DIFF
--- a/src/main/scala/SerialAdapter.scala
+++ b/src/main/scala/SerialAdapter.scala
@@ -12,19 +12,27 @@ import scala.math.min
 case object SerialAdapter {
   val SERIAL_IF_WIDTH = 32
 
-  def connectSimSerial(serial: SerialIO, clock: Clock, reset: Reset) = {
-    val sim = Module(new SimSerial(SERIAL_IF_WIDTH))
-    sim.io.clock := clock
-    sim.io.reset := reset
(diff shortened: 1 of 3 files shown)


src/main/scala/SerialAdapter.scala
```diff
@@ -213,20 +226,9 @@ trait CanHavePeripherySerialModuleImp extends LazyModuleImp {
...
   }
 
-  def connectSimSerial() = {
-    val sim = Module(new SimSerial(SERIAL_IF_WIDTH))
-    sim.io.clock := clock
-    sim.io.reset := reset
-    sim.io.serial <> serial.get
-    sim.io.exit
-  }
+  def connectSimSerial() = serial.map { s => SerialAdapter.connectSimSerial(s, clock, reset) }.getOrElse(false.B)
```
colinschmidt: Why not do the Option[SerialIO] thing like in the block device?

jwright6323: Just to clarify, you mean put the `.map` in the `SerialAdapter` object functions? I can do that.

jwright6323: Did that

PR #71, opened 2020-03-18
Bump sbt version to 2.12.10 for java 13/14

jiegec:


abejgonzalez:
This was done in #107

PR #73, opened 2020-03-26
Fix for latest sifive-blocks uart

tmagik:


zhemao:
Fixed during the last rocket-chip update.

PR #76, opened 2020-04-15
Stage Phase Rocket Chip Bump

davidbiancolin:
See ucb-bar/chipyard#503.

### DIFF
--- a/src/main/scala/TraceIO.scala
+++ b/src/main/scala/TraceIO.scala
@@ -16,7 +16,6 @@ case class TracePortParams(
 
 object TracePortKey extends Field[Option[TracePortParams]](None)
 
-
 case class TracedInstructionWidths(iaddr: Int, insn: Int, cause: Int, tval: Int)
 
 object TracedInstructionWidths {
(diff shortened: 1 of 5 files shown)


src/main/scala/TraceIO.scala
```diff
@@ -83,7 +87,7 @@ object TraceOutputTop {
     new TraceOutputTop(proto.map(t => TracedInstructionWidths(t.head)), proto.map(_.size))
 }
 
-trait CanHaveTraceIO { this: HasTiles =>
+trait CanHaveTraceIO { outer: HasTiles =>
```
zhemao, line 90: Wait, why is it outer instead of this now? You don't make this change for any of the other traits.

src/main/scala/TraceIO.scala
```diff
@@ -93,15 +97,18 @@ trait CanHaveTraceIO { this: HasTiles =>
 }
 
 trait CanHaveTraceIOModuleImp extends LazyModuleImp {
```
jwright6323, line 99: This is probably out of scope for this PR, but should we eventually move to the `InModuleBody` paradigm instead of `ModuleImp`s to mirror rocket-chip?

davidbiancolin: yeah i think so; we can throw out a bunch of extra classes.

jwright6323: OK, filed #77

src/main/scala/TraceIO.scala
davidbiancolin: Yeah this doesn't change anything it's just providing an alias for the self type. I'll put it back.

I don't remember why i changed this, but i think i started rearchitecting this class but then stopped myself half-way without completely reverting all the changes.

PR #78, opened 2020-04-23
Delete Generator/TestHarness and fix ClockUtilTests

zhemao:
This removes the unnecessary Generator/TestHarness, as we can use a separate generator/harness in chipyard for unit-testing all packages.

Also fixes the ClockUtilTests blackboxes so that it can work in VCS. Unfortunately, they won't work in verilator because it doesn't support the delay primitive.

### DIFF
--- a/src/main/resources/testchipip/vsrc/ClockUtil.v
+++ b/src/main/resources/testchipip/vsrc/ClockUtil.v
@@ -49,7 +49,7 @@ module ClockGater (
 
     always @(*) begin
         if (!clockIn) begin
-            qd <= enable;
+            qd = enable;
         end
     end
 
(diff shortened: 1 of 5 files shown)


src/main/resources/testchipip/vsrc/ClockUtil.v
```diff
@@ -49,7 +49,7 @@ module ClockGater (
 
     always @(*) begin
         if (!clockIn) begin
-            qd <= enable;
+            qd = enable;
```
davidbiancolin, line 52: I'm assuming this was intentional?

zhemao: Yeah, this is the correct way, according to @jwright6323 . Verilator complains otherwise.

jwright6323: Yea, @zhemao is cleaning up my mess here. Thanks.

PR #79, opened 2020-04-28
Dromajo Simulation Bridge

abejgonzalez:
Related PR: https://github.com/ucb-bar/chipyard/pull/523

This adds Dromajo as an external "bridge" (similar to FireSim bridges) that can be added to a single core BOOM design. It attaches to the TracePort of the systems. The main modifications are to the TraceIO which now includes an "extended" port that BOOM uses (adds `wdata` to the bundle).


src/main/resources/testchipip/csrc/SimDromajoCosim.cc
```diff
@@ -0,0 +1,114 @@
+#include <vpi_user.h>
+#include <svdpi.h>
+
+#include <stdio.h>
+#include <unistd.h>
+
+#include "dromajo_wrapper.h"
+
+// gotten from elaboration
```
colinschmidt, line 9: Generated during Chipyard elaboration?

src/main/resources/testchipip/csrc/SimDromajoCosim.cc
```diff
@@ -0,0 +1,114 @@
...
+    const char* binary_file)
+{
+    // setup arguments
+    char *local_argv[MAX_ARGS];
+    char local_argc = 0;
+    char mmio_range[MAX_STR_LEN] = "";
+    char plic_params[MAX_STR_LEN] = "";
+    char clint_params[MAX_STR_LEN] = "";
+
+    local_argv[local_argc] = (char*)"./dromajo";
```
colinschmidt, line 29: Is this really the best way to do this? It seems very verbose/messy.

src/main/resources/testchipip/vsrc/SimDromajoCosimBlackBox.v
```diff
@@ -0,0 +1,92 @@
...
+    input [     (XLEN*COMMIT_WIDTH) - 1:0] pc     ,
+    input [(`INST_LEN*COMMIT_WIDTH) - 1:0] inst   ,
+    input [     (XLEN*COMMIT_WIDTH) - 1:0] wdata  ,
+    input [     (XLEN*COMMIT_WIDTH) - 1:0] mstatus,
+    input [          (COMMIT_WIDTH) - 1:0] check  ,
+
+    input           int_xcpt,
+    input [XLEN - 1:0] cause
+);
+    string __dtb_file, __rom_file, __bin_file, __mmio_start, __mmio_end;
```
colinschmidt, line 41: Is there a point to adding these underscores?

src/main/scala/Dromajo.scala
```diff
@@ -0,0 +1,101 @@
...
+ */
+class SimDromajoBridge(insnWidths: TracedInstructionWidths, numInsns: Int) extends Module
+{
+  val io = IO(new Bundle {
+    val trace = Input(new TileTraceIO(insnWidths, numInsns))
+  })
+
+  // constants
+  val xLen = 64
+  val instBits = 32
```
colinschmidt, line 19: This is a requirement of dromajo? Does it get checked against `insnWidths`?

src/main/scala/Dromajo.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  val traces = io.trace.insns
+
+  val dromajo = Module(new SimDromajoCosimBlackBox(numInsns, xLen))
+
+  dromajo.io.clock := clock
+  dromajo.io.reset := reset.asBool
+
+  // TODO: convert to use SInts?
+  def signextPad(in: UInt, padLen: Int): UInt = {
```
colinschmidt, line 30: Seems like you'll need to keep using a function:
https://github.com/chipsalliance/rocket-chip/blob/6ac71323666eee369e326c1dc4e8036a869cea28/src/main/scala/util/package.scala#L104

src/main/scala/Dromajo.scala
```diff
@@ -0,0 +1,101 @@
...
+      Mux(in(in.getWidth - 1) === 1.U, Cat(0.U((padLen - in.getWidth).W) - 1.U, in), in.pad(padLen))
+    }
+  }
+
+  dromajo.io.valid := Cat(traces.map(t => t.valid).reverse)
+  dromajo.io.hartid := 0.U
+  dromajo.io.pc := Cat(traces.map(t => signextPad(t.iaddr, xLen)).reverse)
+  dromajo.io.inst := Cat(traces.map(t => t.insn.pad(instBits)).reverse)
+  dromajo.io.wdata := Cat(traces.map(t => signextPad(t.wdata, xLen)).reverse)
+  dromajo.io.mstatus := 0.U
```
colinschmidt, line 44: Will this work if `mstatus` isn't connected?

src/main/scala/Dromajo.scala
```diff
@@ -0,0 +1,101 @@
...
+class SimDromajoCosimBlackBox(
+  commitWidth: Int,
+  xLen: Int)
+  extends BlackBox(Map(
+    "COMMIT_WIDTH" -> IntParam(commitWidth),
+    "XLEN" -> IntParam(xLen)
+  ))
+  with HasBlackBoxResource
+{
+  val instBits = 32
```
colinschmidt, line 79: If this is required to be the same as the previous constant, you should probably make them reference the same value. Probably in the companion object?

src/main/scala/Generator.scala
```diff
@@ -1,13 +1,7 @@
...
-import chisel3.{Module, Driver}
-import freechips.rocketchip.config.{Parameters, Config}
-import freechips.rocketchip.util.{GeneratorApp, HasGeneratorUtilities, ParsedInputNames}
-import java.io.{File, FileWriter}
-import firrtl.annotations.JsonProtocol
+import freechips.rocketchip.config.Parameters
+import firrtl.options.{StageMain}
+import freechips.rocketchip.system.{RocketChipStage}
 
-object Generator extends GeneratorApp {
```
colinschmidt, line 9: This is already rebased on #78 ?

src/main/scala/TraceIO.scala
```diff
@@ -4,51 +4,105 @@ import chisel3._
...
+    temp.cause := tI.cause
+    temp.tval := tI.tval
+
+    temp
+  }
+
+  def fromVec(tis: Vec[TracedInstruction]): Vec[ExtendedTracedInstruction] = {
+    val tempVec = tis.map(insn => Wire(new ExtendedTracedInstruction()(insn.p)))
+
+    tempVec.zip(tis).foreach({ case (ext, non_ext) =>
```
colinschmidt, line 55: Why can't this apply the previous function?

src/main/resources/testchipip/csrc/SimDromajoCosim.cc
abejgonzalez: Yes. Ill go ahead and update the comment.

src/main/resources/testchipip/csrc/SimDromajoCosim.cc
abejgonzalez: Frankly I don't know. I just quickly mocked this up and didn't spend to much time trying to clean it up since I assumed most people wouldn't touch this code. I agree that it is messy.

src/main/resources/testchipip/vsrc/SimDromajoCosimBlackBox.v
abejgonzalez: I was matching what the block device, dram, simserial, and uart verilog files do for internal signals.

abejgonzalez: With that being said, there is no real meaning other than that they are internal.

src/main/scala/Dromajo.scala
abejgonzalez: Yes. Dromajo is RV64GC. It doesn't check against `insnWidths`

src/main/scala/Dromajo.scala
abejgonzalez: Ill probably switch to this other function then. Good catch.

src/main/scala/Dromajo.scala
abejgonzalez: Yup. Dromajo currently doesn't check against `mstatus`.

src/main/scala/Dromajo.scala
abejgonzalez: Sure.

src/main/resources/testchipip/csrc/SimDromajoCosim.cc
jerryz123: Can't you just initialize the `argv` array in a single expression? https://overiq.com/c-programming-101/array-of-pointers-to-strings-in-c/

src/main/scala/Generator.scala
abejgonzalez: Wait a minute. Should this go into `dev` or into `master`. Did we decide to ditch the `dev` branch on `testchipip`?

src/main/scala/TraceIO.scala
abejgonzalez: I think it can. Ill go ahead and update.

src/main/resources/testchipip/csrc/SimDromajoCosim.cc
abejgonzalez: I think I can do most of it like that. However, I think there needs to be some fanciness done to the MMIO, PLIC, CLINT ranges. Ill mess around and see what works.

src/main/resources/testchipip/csrc/dromajo_wrapper.cc
```diff
@@ -0,0 +1,51 @@
+#include "dromajo_wrapper.h"
+
+dromajo_t::dromajo_t(int argc, char *argv[])
+{
+    //for (int i = 0; i < argc; ++i)
+    //    printf("[DRJ_DEBUG] dromajo arg %d: %s\n", i, argv[i]);
```
colinschmidt, line 6: I don't think we should commit this commented out print.

src/main/scala/TraceIO.scala
```diff
@@ -4,23 +4,54 @@ import chisel3._
...
 
-case class TracedInstructionWidths(iaddr: Int, insn: Int, cause: Int, tval: Int)
+case class TracedInstructionWidths(iaddr: Int, insn: Int, wdata: Int, cause: Int, tval: Int)
 
 object TracedInstructionWidths {
+  def apply(tI: ExtendedTracedInstruction): TracedInstructionWidths =
+    TracedInstructionWidths(tI.iaddr.getWidth, tI.insn.getWidth, tI.wdata.getWidth, tI.cause.getWidth, tI.tval.getWidth)
+
+  // note: the wdata is not 0 here since we can't deal with 0 width wires
   def apply(tI: TracedInstruction): TracedInstructionWidths =
```
davidbiancolin, line 28: [I think] We should be able to deal with an `Option` though, so i'd encourage you to use that in a follow up PR.

src/main/scala/TraceIO.scala
```diff
@@ -77,39 +122,77 @@ class TileTraceIO(val insnWidths: TracedInstructionWidths, val numInsns: Int) ex
...
-// bag of vectors. Each entry is Vec of committed instructions
+// bag of vectors. Each entry is trace associated with a single tile (vector of committed instructions + clock + reset)
 class TraceOutputTop(val widths: Seq[TracedInstructionWidths], val vecSizes: Seq[Int]) extends Bundle {
   val traces = Output(HeterogeneousBag(widths.zip(vecSizes).map({ case (w, n) => new TileTraceIO(w,n) })))
 }
 
 object TraceOutputTop {
-  def apply(proto: Seq[Vec[TracedInstruction]]): TraceOutputTop =
-    new TraceOutputTop(proto.map(t => TracedInstructionWidths(t.head)), proto.map(_.size))
+  def apply(proto: Seq[Vec[TracedInstruction]], protoExt: Seq[Vec[ExtendedTracedInstruction]]): TraceOutputTop =
```
davidbiancolin, line 131: i guess it doesn't cost anything to do it this way but i'm curious why you choose not to simply plumb out a single IO with all possible signals. The bridge/BB specific bundle could have a constructor which accepted that extended type and only connected the pertinent signals.

src/main/scala/TraceIO.scala
abejgonzalez: Huh. I didn't know that. Ill update this after it gets in.

src/main/scala/TraceIO.scala
abejgonzalez: IIRC (since it has been a long time), RC only gives out a `Vec[TracedInstructions]` while BOOM now gives out `Vec[ExtTracedInstructions]`. So to make it easy on me, I just passed in both directly from the traceNexus'es instead of trying to convert the `TracedInstructions` from RC to an extended type. In retrospect, that might be a bit cleaner, I think I can change that in a separate PR just to let this one go in.

abejgonzalez: I think this decision was motivated by me trying to avoid changing RC.

PR #80, opened 2020-04-29
Use rocket-chip's ClockGate API to reduce reduncancy

jwright6323:
This removes the redundant `ClockGater` blackbox in testchipip and replaces it with the `ClockGate` set of APIs from rocket-chip.


jwright6323:
This should be ready to go. Ran `ClockUtilTests` in VCS.

PR #83, opened 2020-05-05
Rc 1.3 bump

colinschmidt:
See: https://github.com/ucb-bar/chipyard/pull/544

### DIFF
--- a/src/main/scala/Ring.scala
+++ b/src/main/scala/Ring.scala
@@ -5,9 +5,9 @@ import chisel3.util._
 
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.subsystem.{SystemBus, SystemBusParams}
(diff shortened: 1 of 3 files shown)


src/main/scala/Ring.scala
```diff
@@ -199,11 +199,22 @@ class TLRingNetwork(
...
+    ring.suggestName(loc.name)
+    context.tlBusWrapperLocationMap += (loc -> ring)
+    ring
+  }
+}
+
+class RingSystemBus(params: SystemBusParams)
     (implicit p: Parameters) extends SystemBus(params) {
-  val system_bus_ring = LazyModule(new TLRingNetwork(buffer))
+  val system_bus_ring = LazyModule(new TLRingNetwork())
```
zhemao, line 215: Instead of getting rid of the `buffer` argument and always using the default defined in TLRingNetword, could you add it to RingSystemBusParams?

PR #87, opened 2020-06-06
Add constructor and parser for a dramsim ini directory from the cmd line

colinschmidt:
This avoid the need to symlink the dramsim_ini directory everywhere.
This also maintains the default behavior if people still want to use that.


colinschmidt:
Thanks for the review waiting for the CI on chipyard before merging.

Issue #88, opened 2020-06-13
SerialAdapter should require minLatency > 0


mwachs5:
There is a state machine in the Serial Adapter that seems to assume that d_valid will not be asserted on the same cycle as a_valid when writing. It can't accept the ack on the same cycle that the data is being written. The state machine won't advance if this is the case, even though that is valid Tile Link behavior.

There should be a `require(serial.minLatency > 0)` e.g. here to detect this case, or the state machine should be corrected to handle it:

https://github.com/ucb-bar/testchipip/blob/2797a6c1c17b90dd415619c3b5c5b731868a3adb/src/main/scala/SerialAdapter.scala#L51

This is my understanding, the state machine won't accept the write ack (by asserting d_ready) unless it's in write ack state:
https://github.com/ucb-bar/testchipip/blob/2797a6c1c17b90dd415619c3b5c5b731868a3adb/src/main/scala/SerialAdapter.scala#L105

But it won't go into write-ack state unless the write is actually accepted:
https://github.com/ucb-bar/testchipip/blob/2797a6c1c17b90dd415619c3b5c5b731868a3adb/src/main/scala/SerialAdapter.scala#L178

But in tile link it is legal to have a_ready = d_ready.

timsnyder:
Thanks for filing this @mwachs5.

> But in tile link it is legal to have a_ready = d_ready.

See also:  "Request Response Ordering" (immediately before Fig 4.3) in [TileLink Spec v1.8](https://sifive.cdn.prismic.io/sifive%2Fcab05224-2df1-4af8-adee-8d9cba3378cd_tilelink-spec-1.8.0.pdf) 

> For example, a designer might be tempted to implement a master interface which holds d ready
LOW while a valid is HIGH in order to delay a concurrent response message until the following
cycle. However, this represents an indefinite delay on Channel D that is not allowed by any of the
forward progress ready rules. Indeed, a TL-UL–conforming slave interface may have connected
d valid and d ready to a valid and a ready respectively. Thus, the non-conforming master
interface has introduced a deadlock. <br> <br> If a master interface cannot deal with receiving a response message on the same cycle as its
request message, then it can instead put a buffer after its Channel D input. The buffer absorbs a
concurrent Channel D response message and presents d ready HIGH until it has been filled.

I've encountered a case where a deadlock does occur.  I'd be happy to submit a PR that at least adds buffering in front of the TL interface if that would be helpful.

zhemao:
[...]

PR #91, opened 2020-06-22
BlockDevice and SerialAdapter require minLatency > 0 (Fix Issue #88)

zhemao:
Let's put this in after the RC bump is concluded.


zhemao:
The L2 agent always adds at least a cycle of delay, so this should be safe. I compiled the DefaultRocketConfig and SimBlockDeviceRocketConfig and the requirement was met.

PR #92, opened 2020-07-10
Handle RC#2521 in DromajoHelper

timsnyder-siv:
https://github.com/chipsalliance/rocket-chip/pull/2521 replaces
   case object BootROMParams extends Field[BootROMParams]
with
   case class BootROMLocated(loc: HierarchicalLocation) extends Field[Option[BootROMParams]](None)

This commit updates DromajoHelper to take a HierarchicalLocation and updates the
lookup of DRAMOJO_RESET_VECTOR adn DRAMOJO_MMIO_START accordingly.

NOTE: you will want to bump RC at the same time that you bump to this ChipYard.  There will also be other changes needed in ChipYard.


abejgonzalez:
This was addressed in #96

PR #93, opened 2020-07-10
Serial writes directly to DRAM for fast program loading

zhemao:
This PR attempts to solve the issue of slow program loading over HTIF by bypassing writes directly to the DRAM model, in a way similar to how FireSim currently functions. The main issue with this approach is that the memory models have to be hosted in SimSerial now, and SimDRAM needs SimSerial to be in the design in order to function properly.

### DIFF
--- a/src/main/resources/testchipip/csrc/SimDRAM.cc
+++ b/src/main/resources/testchipip/csrc/SimDRAM.cc
@@ -4,44 +4,12 @@
 #include <stdint.h>
 
 #include "mm_dramsim2.h"
+#include "serial.h"
 
-int dramsim = -1;
-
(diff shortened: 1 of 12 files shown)


jerryz123:
Doesn't this couple SimSerial to SimDRAM? Does this work when SimAXIMem is used, instead of SimDRAM? What about when the system only has scratchpad memory.

zhemao:
SimSerial can work without SimDRAM, but not vice-versa. I have it set up to not use the fast load feature if nchannels is 0. Though I guess it won't detect that if it's using SimAXIMem.

colinschmidt:
This was fixed in #94

PR #94, opened 2020-07-11
Implement loadmem functionality

zhemao:
This is a less aggressive change than the other one. Instead of attempting to load the executable directly into memory, the user can provide a separate hex file to be loaded into a specified address. This way, SimSerial and SimDRAM don't need to be tied together. 

As a bonus, this also fixes an issue with dramsim_ini_dir not working if there are multiple SimDRAM channels.

### DIFF
--- a/src/main/resources/testchipip/csrc/SimDRAM.cc
+++ b/src/main/resources/testchipip/csrc/SimDRAM.cc
@@ -6,6 +6,9 @@
 #include "mm_dramsim2.h"
 
 int dramsim = -1;
+unsigned long loadmem_addr = 0;
+std::string ini_dir = "dramsim2_ini";
+std::string loadmem_file = "";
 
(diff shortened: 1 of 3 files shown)


jerryz123:
I didn't follow your conversation with Baltazar. Did the `load-mem` code just not work before?

zhemao:
No, it didn't work. It assumed that each channel gets separate data, but with the way mm_t::read and mm_t::write handle addresses, they should actually all get identical data. This greatly simplifies things.

baltazarortiz:
Can confirm it works when given correct arguments! Thank you again for getting this working :)

There isn't an error that I can tell when you provide a filename that doesn't exist or when you tell it to load either starting outside the available memory space *or* give it a file that's bigger than the memory size (it doesn't seem to crash in this case either, which surprised me?).

PR #97, opened 2020-09-05
Clock the PeripherySerialAdapter, BlockDevice from the pbus

jerryz123:


src/main/scala/BlockDevice.scala
```diff
@@ -443,6 +443,11 @@ trait CanHavePeripheryBlockDevice { this: BaseSubsystem =>
     val c = LazyModule(new BlockDeviceController(
       0x10015000, pbus.beatBytes))
 
+    InModuleBody {
+      c.module.clock := pbus.module.clock
```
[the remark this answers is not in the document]
jerryz123: Pretty easy

colinschmidt (approved):
LGTM

src/main/scala/BlockDevice.scala
```diff
@@ -1,20 +1,29 @@
 package testchipip
 
 import chisel3._
+import chisel3.experimental.{IO}
```
davidbiancolin, line 4: This isn't in the main package?(!)

src/main/scala/Util.scala
```diff
@@ -170,3 +171,9 @@ object DecoupledMux {
   def apply[T <: Data](sel: Bool, a: DecoupledIO[T], b: DecoupledIO[T]): DecoupledIO[T] =
     apply(sel, Seq(b, a))
 }
+
+class ClockedIO[T <: Data](gen: T) extends Bundle {
+  val clock = Output(Clock())
+  val bits = gen
+  override def cloneType: this.type = (new ClockedIO(DataMirror.internal.chiselTypeClone[T](gen))).asInstanceOf[this.type]
```
davidbiancolin, line 178: This needs some explanation. Why can't you pass gen through? Related, why can't you rely on autoclonetype?

davidbiancolin (commented):
OK that's really cool. 

Does Rocket Chip not already have wrapper bundles to add clocks and resets?

src/main/scala/BlockDevice.scala
jerryz123: Normally `IO` can only be called within a `Module`. This needs to be imported so we can add IO in a `InModuleBody`

jerryz123:
Rocketchip has `trait Clocked`, but it isn't as convenient to use as this `ClockedIO` thing.

src/main/scala/BlockDevice.scala
davidbiancolin: Ah, yes, thanks.

src/main/scala/Util.scala
jerryz123: > Why can't you pass gen through?

I'm not an expert on this, but I think its because using `gen` as both `val bits` and in `def cloneType` results in two instances of `ClockedIO` using the same instance of `gen`. This results in a chisel `RebindingException`.
Alternatively, 
```
val bits = DataMirror.internal.chiselTypeClone[T](gen)
override def cloneType: this.type = (new ClockedIO(gen)).asInstanceOf[this.type]
``` 
also works.

> why can't you rely on autoclonetype?

That relies on `gen` being accessible, but `(val gen: T)` results in a `AutoCloneTypeException`, with `constructor parameters (gen) have values that are hardware types, which is likely to cause subtle errors. Use chisel types instead:`.

davidbiancolin: You can make gen be a `private val` parameter to fix the autoclonetype problem.

Do you expect gen will sometimes be bound to a hardware type? Do you care about the direction the bound thing has? If the answer is yes, and yes, then you'll need to use a fancier cloneType, otherwise good ole `cloneType` should work for this application.

I'm trying to understand why you need to call to DataMirror.internal (which shouldn't be necessary here). Did chisel complain about having both undirectioned and directioned fields in ClockedIO?

davidbiancolin: I was playing around with this because i was curious. https://scastie.scala-lang.org/qYVPEuV7SfiXfAnh2ACu2g

Using a primitive type breaks everything if you don't wrap it in a direction, but that won't actually happen all that often if ever.

jerryz123: The error is actually due to some interaction with a `cloneType` in the IOCell generators.

https://scastie.scala-lang.org/DChdc5wHQjO6TgGsYjMwvw

davidbiancolin: ooo, that is weird. The assignment of gen to data is transitively creating a real binding for gen... If you remove the data field, or create a new node by cloning gen, it works fine. Pedantically, i think most scala-hygenic way to write this would be to have gen be a function parameter `() => T`, otherwise the direction preserving cloneType is required, which given rocketchip's reliance on chiselTypeClone is fine by me. Writing it this way will lead to the fewest surprises though:

```
class Clocked[T <: Data](private val gen: T) extends Bundle {
  val clock = Output(Clock())
  val data = DataMirror.internal.chiselTypeClone[T](gen)
}
```
https://scastie.scala-lang.org/pKendxz5SOSVoQf2jYxZKg

PR #100, opened 2020-09-21
Add tile-reset control registers

jerryz123:
This is meant to be a generalization of some of the SCRs for controlling tile resets in Eagle/Beagle, implemented as a diplomatic clock node.

There will be a corresponding PR to Chipyard branching off David's https://github.com/ucb-bar/chipyard/pull/676


src/main/scala/TileResetCtrl.scala
```diff
@@ -0,0 +1,64 @@
...
+  val tileResetProviderNode = ClockGroupIdentityNode()
+
+  lazy val module = new LazyModuleImp(this) {
+    val nTiles = p(TilesLocated(InSubsystem)).size
+    require (nTiles <= 4096 / 8)
+    val r_tile_resets = (0 until nTiles).map({ i =>
+      Module(new AsyncResetRegVec(w=1, init=(if (params.initResetHarts.contains(i)) 1 else 0)))
+    })
+    node.regmap((0 until nTiles).map({ i =>
+      (i*8) -> Seq(RegField.rwReg(1, r_tile_resets(i).io)),
```
colinschmidt, line 45: Why not just pack them tightly?

colinschmidt (commented):
Seems reasonable. Is there anything extra that needs to be done to ensure the tile's reset is sync'd with the tile's clock? It seems logical for it to be magically included in diplomatic clocking but I'm not sure if it is.

src/main/scala/TileResetCtrl.scala
jerryz123: Yeah these can just be byte-packed. 
I don't want to do bit-packing, because that makes it slightly harder to enable/disable specific harts.

src/main/scala/TileResetCtrl.scala
```diff
@@ -0,0 +1,64 @@
...
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.devices.tilelink._
+import freechips.rocketchip.regmapper._
+import freechips.rocketchip.subsystem._
+import freechips.rocketchip.util._
+import freechips.rocketchip.tile._
+import freechips.rocketchip.prci.{ClockSinkDomain, ClockGroupIdentityNode}
+
+// initResetHarts: list of hartids which will stay in reset until its reset-ctrl register is cleared
+case class TileResetCtrlParams(initResetHarts: Seq[Int] = Nil, address: BigInt=0x100000, slaveWhere: TLBusWrapperLocation = PBUS)
```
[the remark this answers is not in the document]
jerryz123: `programmableResetHarts` and  `alwaysOnHarts` suggest that harts not included do not receive a reset control register, which is not the case in this implementation.

In this implementation, all harts can have their resets controlled through these registers. This argument only controls the initial state of those registers, which is why I chose the `init` name.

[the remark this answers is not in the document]
colinschmidt: Ok yeah I guess that makes sense. Didn't think about that detail.

src/main/scala/TileResetCtrl.scala
colinschmidt: Ever so slightly but more than 512 harts is unreasonable

jerryz123:
> Is there anything extra that needs to be done to ensure the tile's reset is sync'd with the tile's clock?

In the current system, the clock/reset generator functions in ChipTop are expected to synchronize the resets before wiring them to the clocksource nodes. As a result, this code assumes the incoming tile reset is already synchronized.

I agree that we may want to make the reset-synchronizers diplomatic nodes in the future.

src/main/scala/TileResetCtrl.scala
colinschmidt: It is convenient to be able to bring more than one hart of reset with a single write, but it is also annoying to have to do bit manipulation for a fully bit-packed register (not necessarily for single-bits but bit-packing multi-bit registers has proven annoying).
I think just switching them to byte-packed it the best solution.

colinschmidt (approved):
LGTM

PR #102, opened 2020-09-25
Add config to hang the serial TL mem off the PBus for the no-L2 case.

jerryz123:
For configs with l2banks = 0, there is no MBUS to hang the dummy memory off of. In that case, hang it off the PBUS.


colinschmidt (approved):
LGTM

PR #106, opened 2020-10-09
Add Support for MBUS crossings

davidbiancolin:
This breaks the assumption that the client and slave sides of TLSerial are int the same clock domain, instead, TLSerial couples itself to the client domain, and provides a hook to select the manager-side crossing type. 

As in #103, provides ClockedAndResetIO.

The changes contained within are probably desirable even if we don't like chipyard-side changes.

### DIFF
--- a/src/main/scala/SerialAdapter.scala
+++ b/src/main/scala/SerialAdapter.scala
@@ -231,10 +231,10 @@ case object SerialTLKey extends Field[Option[SerialTLParams]](None)
 
 case class SerialTLAttachParams(
   masterWhere: TLBusWrapperLocation = FBUS,
-  slaveWhere: TLBusWrapperLocation = MBUS
+  slaveWhere: TLBusWrapperLocation = MBUS,
+  slaveCrossingType: ClockCrossingType = AsynchronousCrossing()
 )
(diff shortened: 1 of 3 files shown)


src/main/scala/SerialAdapter.scala
```diff
@@ -261,18 +261,17 @@ trait CanHavePeripheryTLSerial { this: BaseSubsystem =>
...
-    domain.clockNode := manager.fixedClockNode
     val serdesser = domain { LazyModule(new TLSerdesser(
       w = params.width,
       clientPortParams = clientPortParams,
       managerPortParams = managerPortParams
     )) }
     manager.coupleTo(s"port_named_serial_tl_mem") {
-      (serdesser.managerNode
-        := TLBuffer()
+      ((domain.crossIn(serdesser.managerNode)(ValName("TLSerialManagerCrossing")))(p(SerialTLAttachKey).slaveCrossingType)
```
colinschmidt, line 274: Do crossings include a buffer by default? I think we usually want to include one in this path.

colinschmidt (approved):
LGTM.

src/main/scala/SerialAdapter.scala
davidbiancolin: Yes, if the crossing type is "SynchronousCrossing" and not "NoCrossing".

PR #111, opened 2020-12-16
Teach TLTileResetCtrl about TileResetDomain

### LINKED ISSUE #110: update tile reset control registers for TileResetDomain
I'm trying to use https://github.com/ucb-bar/testchipip/commit/60d5da8d4302e968018212444b807e3c98c3c02e with rocket that contains chipsalliance/rocket-chip#2641 .  That PR changes `freechips.rocketchip.tile.TilePRCIDomain` so that it no longer has a `ClockSinkNode`.  It has a `TileResetDomain` that now has the `ClockSinkNode`.

and I'm running into:
```
info] Compiling 20 Scala sources to /home/centos/federation_201214/testchipip/target/scala-2.12/classes ...
[error] /home/centos/federation_201214/testchipip/src/main/scala/TileResetCtrl.scala:49:11: value clockSinkNode is not a member of freechips.rocketchip.tile.TilePRCIDomain[_$1]
[error]         d.clockSinkNode.portParams(0).name.get -> r_tile_resets(i).io.q
[error]           ^
```

I'm starting to look into how to address this but thought I'd file an issue in case @jerryz123, or @hcook or @davidbiancolin  already have looked at this.

timsnyder-siv:
fixes #110 

@jerryz123 and @davidbiancolin I'm going to bump RC in a clone of chipyard and try doing some FireSim builds.  Please let me know if there's something specific that will test that the TLTileResetCtrl is working as intended.

@hcook if you have a chance to take a look at this and see if it makes sense.  It compiles but I have a feeling that I might be missing something here.

### DIFF
--- a/src/main/scala/TileResetCtrl.scala
+++ b/src/main/scala/TileResetCtrl.scala
@@ -46,7 +46,7 @@ class TLTileResetCtrl(w: Int, params: TileResetCtrlParams, tile_prci_domains: Se
     }): _*)
 
     val tileMap = tile_prci_domains.zipWithIndex.map({ case (d, i) =>
-        d.clockSinkNode.portParams(0).name.get -> r_tile_resets(i).io.q
+        d.tile_reset_domain.clockNode.portParams(0).name.get -> r_tile_resets(i).io.q
     })
     (tileResetProviderNode.out zip tileResetProviderNode.in).map { case ((o, _), (i, _)) =>
       (o.member.elements zip i.member.elements).foreach { case ((name, oD), (_, iD)) =>


jerryz123:
Thanks for looking at this. I don't believe we currently have any software which tests this, but for now it should be sufficient to verify that the default configs still work.

timsnyder-siv:
I've confirmed that bumping rocket to latest master hits this problem and then using this gets past it in the scala compile, only to hit compile issues in BOOM.

I'm going to start a CY branch and PR for bumping rocket for the upcoming release.

jerryz123 (approved):

PR #112, opened 2020-12-22
Bootrom support for systems with no S-mode (no MIDELEG)

jerryz123:
This adds support for systems with no supervisor mode support, in which accessing mideleg is an illegal instruction.

@abejgonzalez @a0u could there be an issue with setting mstatus.mie before zeroing mideleg?


src/main/resources/testchipip/bootrom/bootrom.S
```diff
@@ -25,8 +25,13 @@ _hang: // reset vector
   csrw mtvec, a0
   li a0, 8           // MIE or MSIP bit
   csrw mie, a0       // set only MSIP in MIE CSR
-  csrw mideleg, zero // no delegation
   csrs mstatus, a0   // set MIE in mstatus CSR
+  csrr a0, misa      // get MISA mask
+  srai a0, a0, 18    // get 'S' supervisor bit
+  andi a0, a0, 0x1
+  beqz a0, wfi_loop  // only zero mideleg if supported
+  csrw mideleg, zero // no delegation
```
a0u, line 33: Yes, `mstatus.MIE` should be set last to avoid a race.  `mideleg` is not required to be reset to any particular value.

src/main/resources/testchipip/bootrom/bootrom.S
```diff
@@ -25,8 +25,13 @@ _hang: // reset vector
   csrw mtvec, a0
   li a0, 8           // MIE or MSIP bit
   csrw mie, a0       // set only MSIP in MIE CSR
-  csrw mideleg, zero // no delegation
   csrs mstatus, a0   // set MIE in mstatus CSR
+  csrr a0, misa      // get MISA mask
+  srai a0, a0, 18    // get 'S' supervisor bit
```
a0u, line 30: By the way, `gas` also accepts
```suggestion
  srai a0, a0, 'S' - 'A'    // get 'S' supervisor bit
```
which may clarify what that magic number is.

src/main/resources/testchipip/bootrom/bootrom.S
```diff
@@ -21,12 +21,18 @@ interrupt_loop: // trigger MSIP interrupts for other harts so that they also boo
...
-  csrs mstatus, a0   // set MIE in mstatus CSR
+  csrr t0, misa          // get MISA mask
+  srai t0, t0, 'S' - 'A' // get 'S' supervisor bit
+  andi t0, t0, 0x1
+  beqz t0, 1f            // only zero mideleg if supported
+  csrw mideleg, zero     // no delegation
+1:
+  li a0, 8               // MIE or MSIP bit
+  csrs mstatus, a0       // set MIE in mstatus CSR
+  csrw mie, a0           // set only MSIP in MIE CSR
```
a0u, line 34: Just swap these two instructions so that the global interrupt enable bit is set last.  (`mie` is also not guaranteed to be reset.)
```suggestion
  csrw mie, a0           // set only MSIP in MIE CSR
  csrs mstatus, a0       // set MIE in mstatus CSR
```

a0u (approved):
Force-pushed to regenerate the binary images

jerryz123:
Thanks @a0u

PR #113, opened 2020-12-28
TSI Host Widget Revamped

abejgonzalez:
Adapted from #33 

Changes:
- Allows SerDes to handle older TL2 bundles
- Adds basic unit tests for the TSIWidget (only 1 unit test enabled for now)
- Add `TLSinkSetter` widget to change the number of sink bits for a diplomatic TL connection
- Adds a TSI Host Widget - writing to MMIO registers sends a TSI transaction through a SerDes to a DUT. This DUT then can send a response back or can access backing memory over the same SerDes.


src/main/scala/Serdes.scala
```diff
@@ -505,9 +540,9 @@ class TLSerdes(w: Int, params: Seq[TLManagerParameters], beatBytes: Int = 8)
       val inDes = Module(new GenericDeserializer(mergeType, w))
       inDes.io.in <> ser.in
       tl.b.valid := inDes.io.out.valid && inDes.io.out.bits.isB()
-      tl.b.bits := TLMergedBundle.toB(inDes.io.out.bits)
```
jerryz123, line 508: Nit: I wonder if there is a neater way to pass `hasCorruptDenied` around through the module hierarchy. 
Perhaps add it as a Parameters key and use 
```
p.alter({case x: HasCorruptDenied => true})
``` 
to pass the HasCorruptDenied setting into TLSerdesser

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+  // module to take in a decoupled io tsi stream and convert to a TL stream
+  val serialAdapter = LazyModule(new SerialAdapter)
+  // This converts the TL signals given by the serial adapter into a decoupled stream
+  val serdes = LazyModule(new TLSerdesser(
+        w = params.serialIfWidth,
+        clientPortParams =  params.serdesParams.clientPortParams,
+        managerPortParams = params.serdesParams.managerPortParams,
+        hasCorruptDenied = params.serdesParams.hasCorruptDenied))
+
+  // you are sending the TL request outwards... to the serdes manager... then to a serial stream
```
colinschmidt, line 151: ... and then to the real world

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+  serdes.managerNode := TLSourceSetter(params.mmioSourceId) := TLBuffer() := serialAdapter.node
+  // send TL transaction to the memory system on this side
+  // create TL node to connect to outer bus
+  val externalClientNode = serdes.clientNode
+
+  lazy val module = new LazyModuleImp(this) {
+    val io = IO(new Bundle {
+      val adapterSerial = new SerialIO(SerialAdapter.SERIAL_TSI_WIDTH)
+      val serdesSerial =  new SerialIO(params.serialIfWidth)
```
colinschmidt, line 160: Ok now I see why the params weren't being used before but I think it means that the `hostparams.serialIfWidth` should be renamed if its really `hostparams.offChipSerialIfWidth`

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+
+    // connect to the outside world
+    serdesMod.io.ser.in <> io.serdesSerial.in // input decoupled to start serializing (from the target)
+    io.serdesSerial.out <> serdesMod.io.ser.out // output of the SerialAdapter (sends a serialized stream to the target)
+  }
+}
+
+/**
+ * TSIHostWidget to connect the Front End SerVeR which sends TSI to a target TL module
+ * that resides across a SerialIO boundary.
```
colinschmidt, line 178: Maybe add "often off-chip"

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+ *
+ * TX is MMIO -> Queue -> SerialAdapter -> TL -> TLSerdesser Out
+ * RX is TLSerdesser In -> SerialAdapter -> Queue -> MMIO
```
colinschmidt, line 181: I think there should be a mention of the other half of the tlserdesser here.
```
HostTX  is <as above>
Host RX is <as above>
MemoryRX is TLSerdesser In -> ExternalClient -> Host RAM
Memory TX is Host RAM -> ExternalClient -> TLSerdessser Out
```

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+  // create TL node for MMIO
+  val mmioNode = TLIdentityNode()
+  // create TL node to connect to outer bus
+  val externalClientNode = TLIdentityNode()
+
+  // set up async fifos
+  val mmioSink = LazyModule(new TLAsyncCrossingSink)
+  val clientSource = LazyModule(new TLAsyncCrossingSource)
+
+  // setup the TL connection graph
```
colinschmidt, line 204: This comment isn't helpful.
Maybe: "Connect and simplify the MMIO frontend TL"

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+    port.serial_clock := clock
+  }
+
+  /**
+   * Connect the output to the input
+   *
+   * @param port the i/o bundle to connect
+   * @return none
+   */
+  def loopback(port: TSIHostWidgetIO) {
```
colinschmidt, line 332: 👍 cool. Can this create live lock if you write to this widget with its self?

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+   * Tieoff all inputs to the widget
+   *
+   * @param port the i/o bundle interfacing with the widget
+   * @param clock the clock to run the widget on
+   * @return none
+   */
+  def tieoff(port: TSIHostWidgetIO, clock: Clock) {
+    port.serial.in.bits := 0.U
+    port.serial.in.valid := 0.U
+    port.serial.out.ready := 0.U
```
colinschmidt, line 322: Should this be `1.U`?

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
```diff
@@ -0,0 +1,435 @@
...
+                                                                        baseAddress = BigInt(0x0))
+
+  // convert to write, and read expects with the data to patterns to send to pattern pusher
+  val      writePatternSeq = inMMIOWriteSeq.map(WritePattern(params.mmioBaseAddress + TSIHostWidgetCtrlRegs.txQueueOffset,
+                                                             2,
+                                                             _))
+  // get rid of the MSB (fifo full status)
+  val readExpectPatternSeq = outMMIOReadSeq.map(data => ReadExpectPattern(params.mmioBaseAddress + TSIHostWidgetCtrlRegs.rxQueueOffset,
+                                                                          2,
+                                                                          (data << 1) >> 1))
```
colinschmidt, line 156: Does this work? I thought the left shift would add a bit.

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
```diff
@@ -0,0 +1,435 @@
...
+/**
+ * Factory that creates a TSI Seq of words to be used in testing
+ */
+object TSIHelper {
+  // default values for TSI read/write
+  val SAI_CMD_WRITE = BigInt(1)
+  val SAI_CMD_READ = BigInt(0)
+
+  // currently only supports 32b for the word size
+  val dataWidth = 32
```
colinschmidt, line 233: Should this be `SerialAdapter. SERIAL_TSI_WIDTH`?

colinschmidt (commented):
Cool! Do the tests work?
I wonder if this is a good time to just eliminate the `SerialAdapter. SERIAL_TSI_WIDTH` since we haven't had a use case for non-32 bit versions in its entire lifetime. (I think this would make things less confusing).

src/main/scala/tsiHost/TsiHostPeriphery.scala
```diff
@@ -0,0 +1,67 @@
...
+ * Configuration parameter to configure a set of TSI Host Widgets
+ */
+case object PeripheryTSIHostKey extends Field[Seq[TSIHostParams]](Nil)
+
+/**
+ * Trait to create a set of TSI Host Widgets
+ */
+trait HasPeripheryTSIHostWidget { this: BaseSubsystem =>
+  val tsiHostWidgetNodes = p(PeripheryTSIHostKey).map { ps =>
+    // TODO: Measure BW if using sbus instead of pbus
```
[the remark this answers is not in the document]
abejgonzalez: I think this was just a random note to determine the best place to attach the MMIO node. I think it being on the PBUS is probably good enough since we probably aren't looking to do a lot of sustained TSI reqs.

src/main/scala/tsiHost/TsiHostPeriphery.scala
```diff
@@ -0,0 +1,67 @@
...
+  val tsiMemTLNodes = (p(PeripheryTSIHostKey) zip tsiHostWidgetNodes).map { case (params, node) =>
+    val device = new MemoryDevice
+
+    val idBits = 4 // copying the ExtMem
+    val beatBytes = 8 // taken from the MemoryBusKey
+
+    val slaveParams = TLSlavePortParameters.v1(
+      managers = Seq(TLSlaveParameters.v1(
+        address            = Seq(AddressSet(params.targetBaseAddress, params.targetSize - 1)),
+        resources          = device.reg,
```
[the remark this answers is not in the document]
abejgonzalez: This turns into TL memory IOs coming off the `ChipTop`. It is then up to the FPGA implementation work to connect those IOs to the FPGA DRAM. Is that what you are asking about?

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,370 @@
...
+ * Parameters for the SerDes
+ *
+ * @param clientParams client parameters that receives the SerialIO inwards
+ * @param managerParams manager parameters that sends the SerialIO outwards
+ */
+case class TSIHostSerdesParams(
+  clientPortParams: TLMasterPortParameters = TLMasterPortParameters.v1(
+    clients = Seq(TLMasterParameters.v1(
+      name = "tl-tsi-host-serdes",
+      sourceId = IdRange(0, 2)))),
```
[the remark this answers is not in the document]
abejgonzalez: This depends on what the chip SerDes has. Basically, anything in this case class must match the chip.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+  managerPortParams: TLSlavePortParameters = TLSlavePortParameters.v1(
+    managers = Seq(TLSlaveParameters.v1(
+      address = Seq(AddressSet(0, BigInt("FFFFFFFF", 16))), // access everything
```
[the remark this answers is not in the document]
abejgonzalez: Again, this will probably change depending on the chip. These are just "ok defaults" (based on BEAGLE).

src/main/scala/tsiHost/TsiHostPeriphery.scala
```diff
@@ -0,0 +1,67 @@
...
+    // TODO: Measure BW if using sbus instead of pbus
+    val (hostWidget, hostWidgetMemNode) = TLTSIHostWidget.attach(TSIHostWidgetAttachParams(ps, pbus))
+
+    hostWidget
+  }
+
+  val tsiMemTLNodes = (p(PeripheryTSIHostKey) zip tsiHostWidgetNodes).map { case (params, node) =>
+    val device = new MemoryDevice
+
+    val idBits = 4 // copying the ExtMem
```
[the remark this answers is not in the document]
abejgonzalez: IIRC these numbers were added so that you could elaborate the IOs for the FPGA DRAM memory similar to how RC does the AXI4 backing memory and TL MMIO port (https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/subsystem/Ports.scala). 

I think these two parameters need to be set like the `MasterPortParams` since they should really be set by the test chip... not by the bring up platform. I'll push a change.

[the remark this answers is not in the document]
abejgonzalez: Actually on second thought, `idBits` should be specified by the FPGA (since it contains the backing memory) and the `beatBytes` should be specified by the test chip (since the other RC examples do that).

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+    endSinkId = 0,
+    beatBytes = 8),
+  hasCorruptDenied: Boolean = true
+)
+
+
+/**
+ * TSI Host parameter class
+ *
+ * @param serialIfWidth size of the serialIO *out* of the widget (connects TLSerdessers)
```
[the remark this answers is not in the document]
abejgonzalez: Sorry this is the SerDes IO width leaving the widget out to the testchip.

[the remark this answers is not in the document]
abejgonzalez: The "hostSoftwareSerialIfWidth" should never change since the `SerialAdapter` only supports 32b

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+ * @param txQueueEntries size of the queue for sending TSI requests
+ * @param rxQueueEntries size of the queue for receiving TSI responses
+ * @param mmioBaseAddress start address of the MMIO registers
+ */
+case class TSIHostParams(
+  serialIfWidth: Int = 32,
+  txQueueEntries: Int = 16,
+  rxQueueEntries: Int = 16,
+  mmioBaseAddress: BigInt = BigInt("10017000", 16),
+  mmioSourceId: Int = 2,
```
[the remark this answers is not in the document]
abejgonzalez: Actually, this `mmioSourceId` sets the manager side of the SerDes so it is decoupled from the client `sourceId`. And no this is derived from the chip.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * I/O bundle to connect to the MMIO frontend class
+ */
+trait TLTSIHostMMIOFrontendBundle {
+  implicit val p: Parameters
+
+  val params: TSIHostParams
+
+  val serial = new SerialIO(SerialAdapter.SERIAL_TSI_WIDTH)
```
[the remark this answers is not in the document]
abejgonzalez: Sorry this is a bit confusing... the `serialIfWidth` is the offchip LBWIF size while the `SerialAdapter.SERIAL_TSI_WIDTH` is the size of the data needed for the `SerialAdapter`. They can be different sizes.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+ */
+trait TLTSIHostMMIOFrontendModule extends HasRegMap {
+  implicit val p: Parameters
+
+  val io: TLTSIHostMMIOFrontendBundle
+
+  val params: TSIHostParams
+
+  // tsi format input/output queues
+  val txQueue = Module(new Queue(UInt(SerialAdapter.SERIAL_TSI_WIDTH.W), params.txQueueEntries)) // where is the queue being dequeued (to the SerialAdapter)
```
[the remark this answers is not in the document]
abejgonzalez: See prior response.

src/main/scala/tsiHost/TsiHostWidget.scala
```diff
@@ -0,0 +1,335 @@
...
+  val rxQueue = Module(new Queue(UInt(SerialAdapter.SERIAL_TSI_WIDTH.W), params.rxQueueEntries)) // where is the queue being enqueued (from the SerialAdapter)
+
+  // connect queues to the backend
+  io.serial.out <> txQueue.io.deq
+  rxQueue.io.enq <> io.serial.in
+
+  // memory mapped registers and connections to the queues
+  regmap(
+    TSIHostWidgetCtrlRegs.txQueueOffset -> RegFieldGroup("txdata", Some("Transmit data"),
+                                           NonBlockingEnqueue(txQueue.io.enq, 64)),
```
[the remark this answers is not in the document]
abejgonzalez: I never thought of RV32... I don't know how MMIO operations operate when the core type changes. I can just require RV32.

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: Yup. I renamed this.

src/main/scala/tsiHost/TsiHostWidget.scala
abejgonzalez: Maybe. I haven't tried it. Frankly, I don't use this in the current unit tests.

src/main/scala/tsiHost/TsiHostWidgetUnitTest.scala
abejgonzalez: Actually, I think this does nothing since the data is 32b to begin with. I'll remove this logic since the issue is actually  that you can't deal with non-blocking queues in this unit test

src/main/scala/Serdes.scala
abejgonzalez: I noticed that... but I think this is fine for now. That should be easy to do later if need be.

abejgonzalez:
> Cool! Do the tests work?

One does! The other larger one (the full unit test) doesn't because it can't handle non-blocking enqueues and dequeues yet. Ill leave this to future work since the other unit test tests the entire backed (where there will more likely be more issues).

> I wonder if this is a good time to just eliminate the `SerialAdapter. SERIAL_TSI_WIDTH` since we haven't had a use case for non-32 bit versions in its entire lifetime. (I think this would make things less confusing).

Im not sure I follow, the `SERIAL_TSI_WIDTH` is a hardcoded value? What do you mean "remove".

colinschmidt:
By remove I meant just make that parameter invisible but I guess it's somewhat pervasive and so needs to be referenced in many places so keeping the magic number in a define is probably best.

colinschmidt (approved):
Looks good to me.
I hope there are docs on how to use this in chipyard. I know it has caused several of us issues over time, and while this helps by pushing more of the infrastructure upstream it doesn't seem particularly safer than the current state(other than having tests for this widget).
It would be a cool feature in chipyard to be able to elaborate both the FPGA and DUT side at once to get the diplomatic safety and be able to run an "integration test".

abejgonzalez:
> I hope there are docs on how to use this in chipyard. I know it has caused several of us issues over time, and while this helps by pushing more of the infrastructure upstream it doesn't seem particularly safer than the current state(other than having tests for this widget).

There are no docs at the moment. The main place this is "documented" is in a fake bringup config (where this is used). Ill add an issue to add some documentation on how to use this. I know this is problematic but for now pushing this upstream will be easier to manage for the couple chips coming out (rather than trying to find the proper branch, updating it, verifying it works, etc).

> It would be a cool feature in chipyard to be able to elaborate both the FPGA and DUT side at once to get the diplomatic safety and be able to run an "integration test".

I think @jerryz123 mentioned this at some point... however, I think this is a stretch goal for now since we are limited in resources atm. Again Ill file an issue.

jerryz123:
Is this going into the release? If so lets get this merged, then bump testchipip.

jerryz123:
Since chipyard dev currently references this branch (tsktsk), I'm merging this now.

abejgonzalez:
> Since chipyard dev currently references this branch (tsktsk), I'm merging this now.

Oops. Thanks!

PR #114, opened 2020-12-29
Add CI

abejgonzalez:
This PR adds CI similar to https://github.com/riscv-boom/riscv-boom and https://github.com/ucb-bar/gemmini but only runs the synthesizable unit tests.


.circleci/defaults.sh
```diff
@@ -0,0 +1,126 @@
...
+    # NOTE: for some reason without re-invoking screen after this command it breaks
+    # NOTE: AND, screen -list returns false by default (even if there is a session running)
+    run_impl $AWS_SERVER "screen -S CI-$1-SESSION -dm $REMOTE_AWS_WORK_DIR/$3; screen -list || true"
+}
+
+# remove the work dir on the aws server
+clean_aws () {
+    run_aws "rm -rf $REMOTE_AWS_WORK_DIR"
+}
+
```
jerryz123, line 93: This AWS stuff isn't use here, right?
Its up to you if you want to keep it. I don't foresee us ever needing to use this feature in testchipip

.circleci/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+These are the scripts that Circle CI uses to run the tests during a PR.
+
+Note: This uses the Chipyard `ucb-bar/chipyard:dev` branch to run CI (or other similar commits).
+
+Note: CircleCI env. var. refers to adding a environment variable by navigating to the "Project Settings" in the web UI and adding
+an environment variable in the "Environment Variables" tab.
+
+Note: This uses `$SERVER` and `$CI_DIR` which is given in the CircleCI env. var. setup to specify a server to build on.
+To change these variables you must change the project settings of CircleCI.
```
jerryz123, line 12: Since these are just unit tests, is it necessary to build on a remote server? 
I think minimizing the amount of stuff which depends on that remote would reduce the pain if that server ever needs to be updated.

abejgonzalez: For now, I think that is fine, I was just reusing the infrastructure so that people could extend this in the future.

.circleci/defaults.sh
abejgonzalez: I'll remove it. I forgot to delete this.

jerryz123 (approved):

PR #115, opened 2021-01-05
Add config fragment to set SerialTL width

jerryz123:


abejgonzalez:
This PR doesn't have the CI stuff integrated... but that's fine since this should be innocuous.

PR #120, opened 2021-02-23
Added JTAG-to-memory-mapped bus (TL & AXI4) master bridge.

milovanovic:
This PR adds JTAG-to-memory-mapped bus master bridge, right now only for TL and AXI4, but could be potentially extended to other buses like AHB and/or APB. The main idea is to be able to test (in hardware) some memory-mapped peripherals (e.g., a `DspBlock` from [dsptools](https://github.com/ucb-bar/dsptools)) by reading from and writing to their status and configuration registers without having to invoke a complete processor core like Rocket. It is perhaps not intended to be included as is, so feel free to give suggestions how it can be better tailored and streamlined into Chipyard. Unit tests are still missing but can be written if necessary. Could also happen that this repository is not the best place for this block (maybe dsptools would be a better choice). Portions of code are reused from [chisel-jtag](https://github.com/ucb-art/chisel-jtag). Feel free to change the license yourself (or suggest the license change) to whatever you believe is the most suitable one.


colinschmidt:
I haven't read through this completely yet but it seems very cool. Thanks for submitting it!

The one big question I have is what you would expect the operation to be like if the user wanted to include this functionality and the version of jtag that uses a full processor? Can they share the same pins? Can they operate at the same time?

milovanovic:
Well, having in mind the Rocket chip [diagram](https://github.com/ucb-bar/chipyard/blob/master/docs/_static/images/rocketchip-diagram.png), and under the assumption that its debug unit supports all the commands as the proposed jtag2mm bridge, and that it can propagate those commands down to `pbus`, perhaps it wouldn't make sense to duplicate the functionality and have them both operational at the same time (neither to include them both in a design because you would then end up having an additional `pbus` master). Hence, it might even be meaningful to forbid their coexistence.

As I've said, the goal was just to provide a convenient and easy way of driving and testing memory-mapped peripherals during the development phase without instantiating a processor inside the design. Therefore, I view it as some kind of a side utility.

jerryz123:
Do you happen to have an example of how this is integrated in a chip context? Are you building some Rocketchip design with no Tiles, and only peripherals?

milovanovic:
> Do you happen to have an example of how this is integrated in a chip context?

I'm not sure whether you thought on this, but [here](https://github.com/milovanovic/chipyard/tree/JTAG2MMTest) you can find a simple example of a multiplexer wrapped inside `DspBlock` with appropriate tests inside the Chipyard context.

> Are you building some Rocketchip design with no Tiles, and only peripherals?

Actually, the background story is the following: we were participating in the free SKY130 shuttle with two designs ([5](https://efabless.com/projects/5) & [36](https://efabless.com/projects/36)) written in Chisel. Initially we wanted to use Chipyard, but they already provided a [small RISC-V core](https://github.com/cliffordwolf/picorv32) in the harness part of the chip so it was meaningless for us to add another one. However, the problem was that they offered only the Wishbone Bus interconnect, hence we had to use a [third-party bridge](https://github.com/ZipCPU/wb2axip) to connect our AXI4 accelerators. Since I wasn't sure that the other part will work, to mitigate risk, we decided to use an additional AXI4 bus master that we could control independently from a JTAG port. That's why we made [JTAG2MM](https://github.com/milovanovic/jtag2mm) which we now use on a daily basis to test our `DspBlock` peripherals. It's a really handy tool especially for quick FPGA testing where you don't have to place a complete Rocket core just to test an accelerator. I thought it would be good to have it somewhere in the Chipyard also, but I wasn't sure what's the right place for it, `dsptools` or `testchipip`. Therefore, you are absolutely right, in principle it seems as if we are building a Rocketchip design with no Tiles, just the peripherals, that we want to use for quick in-hardware accelerator tests.

jerryz123:
Just FYI, we haven't forgotten about this. Before we merge, we should add an example in Chipyard demonstrating the integration. Haven't found the cycles to look at that yet.

milovanovic:
I've been talking with @colinschmidt today and perhaps the quickest illustrative demo example would be to tie some of the [DspBlocks](https://github.com/ucb-bar/chipyard/blob/master/generators/chipyard/src/main/scala/example/dsptools/DspBlocks.scala) that already reside in Chipyard and just write to and read from their memory-mapped registers with the proposed JTAG bridge. Of course, it's up to @jerryz123 to decide on that, but this would be my proposal.

jerryz123:
That seems very reasonable to me, it would also demonstrate DSP tools usage

PR #121, opened 2021-03-03
AXI Memory Over TL Serial Link

abejgonzalez:
This setup matches closely with the current Berkeley bringup setup where the AXI memory interface resides over the serial link and is in a different clock domain. 

- Adds new `connectHarnessMultiClockAXIRAM` and `MultiClockSerialAXIRAM` functions/modules that have AXI memory interface over TL serial link
- Defaults to using async crossing
- Expands `SerialTLParams` to keep track of AXI clock freq. (used in IO/HarnessBinder to specify clock source and correctly give `SimDRAM` frequency)
- Add config. fragments to change AXI clock freq. and modify BlockDevice attach locations


jerryz123:
Niceeeee. 
Since this should be strictly better than the Harness TLRAM, can we replace the chipyard examples (if any) that use the TLRAM with this implementation?

abejgonzalez:
While this is strictly better than those examples... there is one caveat... passing an extra clock/reset wire from the SoC over to this harness module. This is needed for FireSim to work (since all clocks need to come from the SoC... unless you want the clock bridge to be wired weirdly to both the harness and the SoC - though both reside in the "target" domain)

colinschmidt (commented):
Seems pretty reasonable to me.

src/main/scala/SerialAdapter.scala
```diff
@@ -341,3 +382,117 @@ class SerialRAM(
...
+      port.bits <> m
+      port.clock := memClkRstSource.out.head._1.clock
+      port.reset := memClkRstSource.out.head._1.reset
+      port
+    })
+    ports
+  }
+
+  lazy val module = new LazyModuleImp(this) {
+    val io = IO(new Bundle {
```
jerryz123, line 474: Make this a LazyRawModuleImp if the implicit clock/reset aren't used

src/main/scala/SerialAdapter.scala
```diff
@@ -32,6 +38,29 @@ case object SerialAdapter {
...
+    val ram = LazyModule(new MultiClockSerialAXIRAM(
+      p(SerialTLKey).get.width,
+      p(SerialTLKey).get.memParams,
+      p(SerialTLKey).get.axiMemOverSerialTLParams.get,
+      managerEdge = serdesser.managerNode.edges.in(0),
+      clientEdge = serdesser.clientNode.edges.out(0)
+    ))
+
+    withClockAndReset(serial_port.clock, reset) {
+      val module = Module(ram.module)
```
jerryz123, line 53: Can you look at my changes in https://github.com/ucb-bar/testchipip/pull/124? 
I think its better to clock this module by the harnessClock, and have AsyncQueues do the crossing, rather than giving this module the serial clock directly.

Using AsyncQueues is also safer for a chip with a more realistic clock/reset setup, since there is no guarantee on synchronization between harness-level reset and the reset for the on-chip serdes.

src/main/scala/SerialAdapter.scala
```diff
@@ -341,3 +382,117 @@ class SerialRAM(
...
+      :*= AXI4UserYanker()
+      :*= AXI4IdIndexer(idBits)
+      :*= TLToAXI4()
+      :*= TLWidthWidget(memBusParams.beatBytes)
+      :*= memXbar)
+
+    InModuleBody { memNode.makeIOs() }
+  }
+
+  val mem_axi4 = InModuleBody {
```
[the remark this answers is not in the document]
abejgonzalez: The `memNode` resides in the `memClkRstDomain`, so when `makeIOs` is called it creates the IOs out of that domain (that domain turns into a module). The `mem_axi4` is needed to connect the IOs from that clock domain module to everything else.

src/main/scala/SerialAdapter.scala
abejgonzalez: This module still uses the implicit clock/reset for the `SerialAdapter`/`TLSerdesser`/`TLBuffer`

src/main/scala/SerialAdapter.scala
abejgonzalez: Sure. I think the cleanest way to integrate those changes is to merge that in Chipyard 1st then this PR will get updated with those changes.

src/main/scala/SerialAdapter.scala
```diff
@@ -257,7 +286,30 @@ class SimSerial(w: Int) extends BlackBox with HasBlackBoxResource {
...
+  crossingType: ClockCrossingType = AsynchronousCrossing() // Default to async crossing
+)
+case class AXIMemOverSerialTLClockParams(
+  axiClockParams: Option[AXIClockParams] = Some(AXIClockParams()) // if set, axi port in different clk domain
+) {
+  def getMemFrequency(system: HasTileLinkLocations)(implicit p: Parameters): Double = {
+    axiClockParams match {
+      case Some(clkParams) => clkParams.clockFreqMHz * (1000 * 1000)
+      case None => {
+        // get freq. from what the master of the serial link specifies
```
colinschmidt, line 300: Isn't this getting the frequency from what the serial link masters rather than what masters the serial link?

src/main/scala/TileResetCtrl.scala
```diff
@@ -61,7 +61,8 @@ class TLTileResetCtrl(w: Int, params: TileResetCtrlParams, tile_prci_domains: Se
             // clocked to the bus this is attached to, not the clock in this
             // clock bundle. We expect a ClockGroupResetSynchronizer downstream
             // to synchronize the resets
-            oD.reset := r.asBool.asAsyncReset
+            // Also, this or enforces that the tiles come out of reset after the reset of the system
+            oD.reset := (r.asBool || asyncResetSinkNode.in.head._1.reset.asBool).asAsyncReset
```
colinschmidt, line 65: How does this enforce your claim? The register is reset by the asyncReset and then can only be changed by something else in the system (which would need to be out of reset).
If you want to enforce a delay you should actually put a delay on the reset IMO.

colinschmidt (approved):
Everything looks good except for the reset change at the end.

src/main/scala/TileResetCtrl.scala
jerryz123: This async reset reg isn't necessarily init=1. In the case where init=0, we still need to have the time receive the async reset.

colinschmidt: Ok. I see now. In that past this hasn't been an issue since the `init=0` or always on blocks just directly receive the off-chip reset without a register to control them.

abejgonzalez: Since the SCR reg. is by default set to `0`, then the core is out of reset before the rest of the system in the current setup. By adding the `||` you enforce that the "global" async reset is 1st de-asserted before the async SCR register has any effect (at that point you toggle the tile reset). This isn't an issue of delay because the initial state of the SCR register is `0`.

src/main/scala/SerialAdapter.scala
abejgonzalez: Oh yea this comment is off. The serial link masters the FBUS and is slave to the MBUS. The code is still fine though.

PR #123, opened 2021-03-12
Add ClockDividerAndMux util module

jerryz123:


abejgonzalez (approved):
LGTM

colinschmidt (approved):
LGTM, with just one question.

src/main/scala/ClockUtil.scala
```diff
@@ -164,6 +164,26 @@ class ClockDivider(width: Int) extends Module {
 
 }
 
+class ClockDivideAndMux(width: Int, depth: Int = 3, genClockGate: () => ClockGate) extends Module {
```
abejgonzalez, line 167: Do you care about stopping the clock? Why not just make this be the divider on `divisor = [1, N)` and the passthrough on `divisor = 0`? No subtract operation needed?

jerryz123: Good point. I actually prefer this way. If we want to stop the clock, we can use the existing clock gating stuff.

src/main/scala/ClockUtil.scala
```diff
@@ -84,11 +84,11 @@ class ClockMutexMux(val n: Int, depth: Int, genClockGate: () => ClockGate) exten
     val io = IO(new Bundle {
         val clocksIn = Input(Vec(n, Clock()))
         val clockOut = Output(Clock())
-        val resetAsync = Input(Bool())
+        val resetAsync = Input(Reset())
```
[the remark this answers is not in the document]
jerryz123: Fixed

abejgonzalez (approved):
LGTM. I approved the other version anyways... I figured you wanted to keep the "clock stop" functionality.

jerryz123:
Clock stop functionality can always be added with a clock gate on the output clock of this block, which I think is a neater way to do this.

PR #124, opened 2021-03-15
Add support for adding AsyncQueues when sending serial off-chip

jerryz123:
Also avoid clocking the harness-side serdes with the onchip serial clock. Use an async queue instead


src/main/scala/SerialAdapter.scala
```diff
@@ -15,18 +15,47 @@ import scala.math.min
...
+      in_queue.io.enq_clock := clock
+      in_queue.io.enq_reset := reset.asBool
+      val crossed = Wire(new SerialIO(w))
+      in_queue.io.enq <> crossed.in
+      crossed.out <> out_queue.io.deq
+      crossed
+    }
+  }
+
+  def asyncQueue(port: SerialIO, clock: Clock, reset: Reset): SerialIO = {
```
colinschmidt, line 41: If the inner method is going to be used on the test harness side as an actual asyncqueue  I think we should rename this helper method so people aren't confused when it instantiates a synchronous (but async reset) queue.

colinschmidt (approved):
LGTM, should make sure the harness side uses the asyncQueue helper from here.

abejgonzalez (approved):
LGTM

PR #125, opened 2021-03-22
Add config fragment to control initialization of tiles in reset

jerryz123:


src/main/scala/Configs.scala
```diff
@@ -95,3 +95,8 @@ class WithSerialTLMem(
     isMemoryDevice = isMainMemory
   )}
 })
+
+class WithTilesStartInReset(harts: Int*) extends Config((site, here, up) => {
+  case TileResetCtrlKey => up(TileResetCtrlKey, site).copy(initResetHarts = up(TileResetCtrlKey, site).initResetHarts ++ harts)
```
colinschmidt, line 100: I'm wondering if there should be a warning or error if you set the same hart twice?
From the code I understand that the operation would be idempotent but potentially the user has made a mistake and meant to include a different hart not currently in the list.

colinschmidt (approved):
LGTM. A small question but I don't care how its resolved.

src/main/scala/Configs.scala
jerryz123: Perhaps. I'd like to see this functionality used more before evaluating what the most intuitive behavior is.

PR #128, opened 2021-03-30
Test chip features

jerryz123:
A set of useful features for testchip

- The `+init_write=<addr>:<val>` argument lets the user write to arbitrary SCRs before hart0 leaves the bootrom. Useful for setting up clocks, putting things in and out of reset, etc. You can string together an arbitrary sequence of `init_write` flags to set up the SCRs in any way you want.
- The bootrom now references a value stored in a `BootAddrReg` memory-mapped register, instead of always sending the harts to `DRAM_BASE`. The default is still `DRAM_BASE`
- The new `CustomBootPin` functionality exposes a ChipTop pin that, when asserted, writes some custom address to the `BootAddrReg`, and then writes MSIP for hart0. This lets us simulate untethered boot, since we can redirect hart0 to secondary boot media without relying on TSI for anything. 
- The `+no_hart0_msip` argument prevents the tsi widget from breaking hart0 out of the bootrom by writing MSIP. When coupled with the `CustomBootPin` functionality, this lets us simulate untethered boot
- TileResetCtrl control registers are restored to word-packed, as opposed to byte-packed. There is a longstanding bug in upstream riscv-fesvr where `read_chunk` and `write_chunk` does not work with sub-word accesses. https://github.com/riscv/riscv-isa-sim/blob/21684fd9b073cf9bd8f8d23cfc5f94ce361f170c/fesvr/tsi.cc#L55 . Making the ResetCtrl registers word-packed lets them be manipulated through TSI, which is useful.


src/main/resources/testchipip/csrc/testchip_tsi.cc
```diff
@@ -3,10 +3,22 @@
...
+  init_writes = std::vector<std::pair<uint64_t, uint32_t>>();
+
   std::vector<std::string> args(argv + 1, argv + argc);
   for (auto& arg : args) {
     if (arg.find("+loadmem=") == 0)
       has_loadmem = can_have_loadmem;
+    if (arg.find("+init_write=") == 0) {
+      auto d = arg.find(":");
+      if (d == std::string::npos) {
+	throw std::invalid_argument("Improperly formatted +init_write argument");
```
abejgonzalez, line 15:
```suggestion
	  throw std::invalid_argument("Improperly formatted +init_write argument");
```

src/main/resources/testchipip/csrc/testchip_tsi.cc
```diff
@@ -3,10 +3,22 @@
...
+      }
+      uint64_t addr = strtoull(arg.substr(12, d - 12).c_str(), 0, 16);
+      uint32_t val = strtoull(arg.substr(d + 1).c_str(), 0, 16);
```
abejgonzalez, line 18: This enforces that you can only put hex numbers, that's fine but can that be documented somewhere or maybe have an extra check to ensure there is a `0x` prefix.

src/main/resources/testchipip/csrc/testchip_tsi.cc
abejgonzalez, line 15:
```suggestion
	  throw std::invalid_argument("Improperly formatted +init_write argument");
```

abejgonzalez (changes requested):
I think this looks nice. The one assumption that I think you are making is that you will write to all SCRs in batch. In my test chip I write SCRs in phases (I have some scripts that write SCRs to reset... wait a bit... set clocks.... wait a bit / calc stuff... set more stuff). However, this seems like a useful 1st step for simulation purposes.

jerryz123:
> I think this looks nice. The one assumption that I think you are making is that you will write to all SCRs in batch. 

Yeah. For anything more complicated this simple solution breaks down. But it is useful for simulation in this form

src/main/scala/CustomBootPin.scala
```diff
@@ -0,0 +1,87 @@
...
+  case CustomBootPinKey => up(CustomBootPinKey, site).map(p => p.copy(customBootAddress = address))
+})
+
+trait CanHavePeripheryCustomBootPin { this: BaseSubsystem =>
+  val custom_boot_pin = p(CustomBootPinKey).map { params =>
+    val tlbus = locateTLBusWrapper(params.masterWhere)
+    val clientParams = TLMasterPortParameters.v1(
+      clients = Seq(TLMasterParameters.v1(
+        name = "custom-boot",
+        sourceId = IdRange(0, 1)
```
colinschmidt, line 33: I believe your state machine requires `minLatency = 1` as it can't handle combo A->D

src/main/scala/CustomBootPin.scala
```diff
@@ -0,0 +1,87 @@
...
+              lgSize = 2.U,
+              data = params.customBootAddress.U
+            )._2
+            when (tl.a.fire()) { state := waiting_bootaddr_reg_d }
+          }
+          is (waiting_bootaddr_reg_d) { when (tl.d.fire()) { state := waiting_msip_a } }
+          is (waiting_msip_a) {
+            tl.a.valid := true.B
+            tl.a.bits := edge.Put(
+              toAddress = p(CLINTKey).get.baseAddress.U, // msip for hart0
```
colinschmidt, line 67: We might want to use the helper function to fill this in.
```
toAddress = p(CLINTKey).get.baseAddress.U + CLINTConts.msip(0),
lgSize = log2(CLINTConsts.msipBytes)
```

src/main/scala/CustomBootPin.scala
```diff
@@ -0,0 +1,87 @@
...
+              fromSource = 0.U,
+              lgSize = 2.U,
+              data = 1.U
+            )._2
+            when (tl.a.fire()) { state := waiting_msip_d }
+          }
+          is (waiting_msip_d) { when (tl.d.fire()) { state := inactive } }
+          is (dead) { when (!custom_boot) { state := inactive } }
```
colinschmidt, line 75: I don't understand this state? How does it enter dead? Why is it not just always in inactive?

src/main/scala/TileResetCtrl.scala
```diff
@@ -38,14 +38,14 @@ class TLTileResetCtrl(w: Int, params: TileResetCtrlParams, tile_prci_domains: Se
...
-    require (nTiles <= 4096)
+    require (nTiles <= 4096 / 4)
     val r_tile_resets = (0 until nTiles).map({ i =>
       withReset (asyncResetSinkNode.in.head._1.reset) {
         Module(new AsyncResetRegVec(w=1, init=(if (params.initResetHarts.contains(i)) 1 else 0)))
       }
     })
     node.regmap((0 until nTiles).map({ i =>
-      i -> Seq(RegField.rwReg(1, r_tile_resets(i).io)),
+      i * 4 -> Seq(RegField.rwReg(1, r_tile_resets(i).io)),
```
colinschmidt, line 48: This change still seems silly to me.
The only benefit seems to be the user doesn't need to know that state of the cores to affect a change, i.e. they can just write a 0 or 1 to a specific word without reading first or keeping of track of which cores are in reset.
The down side is the user cannot bring multiple cores out of reset at once. This can be problematic if the user has cores that are designed to come up together.

colinschmidt (changes requested):
I think the `init_writes` feature is useful but as Abe mentioned for an actual chip it is probably insufficient and people would end up writing their own `my_chip_tsi_t` class or something more custom like that to solve this issue. It is worth at least mentioning that in the documentation for this new feature on what to do when this in insufficient.

The `CustomBootPin` feature seems to be missing the part of self-boot where the bootrom is updated to read from some off-chip storage before jumping to it. Is the idea that one would use yet-another-boot-loader(YABL) in some other ROM/RAM for the reading from off-chip storage?

My other question about CustomBootPin is why do we choose to make this a hardware state machine vs the traditional way we have done this in bootrom SW?

And finally compressing the reset regs doesn't seem very useful to me and removes some functionality.

src/main/scala/BootAddrReg.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  val tlbus = locateTLBusWrapper(params.slaveWhere)
+
+  val device = new SimpleDevice("boot-address-reg", Nil)
+
+  tlbus {
+    val node = TLRegisterNode(Seq(AddressSet(params.bootRegAddress, 4096-1)), device, "reg/control", beatBytes=tlbus.beatBytes)
+    tlbus.toVariableWidthSlave(Some("boot-address-reg")) { node := TLBuffer() }
+    InModuleBody {
+      val bootAddrReg = RegInit(params.defaultBootAddress.U(32.W)) // 32bits wide for uniform behavior between 32bit and 64bit systems
```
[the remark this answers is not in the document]
jerryz123: A potential future where someone's 64-bit core jumps correctly to the alternate boot ROM/RAM, but their 32 bit core doesn't , and they have a hard time figuring out why.

Boot media should probably be mapped to low addresses anyways, this helps enforce that.

src/main/scala/TileResetCtrl.scala
jerryz123: The fact that byte-wide writes through TSI don't work is the real motivation for this change. This just seems safer for now, as its more intuitive to do a series of word-writes, instead of doing byte-mask manipulation.

Also if some cores are designed to come up together with reset, then some other widget should be used instead of this one. Even if we byte-packed them, this widget wouldn't support bringing them out of reset together if their hartids are more than 8 apart, or not aligned properly.

src/main/scala/BootAddrReg.scala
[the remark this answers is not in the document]
jerryz123: Actually since it seems like a system with both 64bit and 32bit can't really be supported due to shared bootrom, I'll just make this 64-bits wide

src/main/scala/BootAddrReg.scala
```diff
@@ -0,0 +1,38 @@
...
+trait HasPeripheryBootAddrReg { this: BaseSubsystem =>
+  val params = p(BootAddrRegKey)
+
+  val tlbus = locateTLBusWrapper(params.slaveWhere)
+
+  val device = new SimpleDevice("boot-address-reg", Nil)
+
+  tlbus {
+    val node = TLRegisterNode(Seq(AddressSet(params.bootRegAddress, 4096-1)), device, "reg/control", beatBytes=tlbus.beatBytes)
+    tlbus.toVariableWidthSlave(Some("boot-address-reg")) { node := TLBuffer() }
```
abejgonzalez, line 32: Stupid question, do we need a buffer here? Can we just directly wire the register to the PBUS? I think this is simple enough to not need any buffering.

src/main/scala/BootAddrReg.scala
```diff
@@ -0,0 +1,38 @@
...
+
+trait HasPeripheryBootAddrReg { this: BaseSubsystem =>
+  val params = p(BootAddrRegKey)
+
+  val tlbus = locateTLBusWrapper(params.slaveWhere)
+
+  val device = new SimpleDevice("boot-address-reg", Nil)
+
+  tlbus {
+    val node = TLRegisterNode(Seq(AddressSet(params.bootRegAddress, 4096-1)), device, "reg/control", beatBytes=tlbus.beatBytes)
```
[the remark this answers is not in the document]
jerryz123: Using precise mask size makes the supportsAcquire checks in the TLBs more costly. All existing tilelink devices uses a mask size of 4096 or greater.

src/main/scala/CustomBootPin.scala
```diff
@@ -0,0 +1,86 @@
...
+import freechips.rocketchip.prci._
+
+case class CustomBootPinParams(
+  customBootAddress: BigInt = 0x80000000L, // Default is DRAM_BASE
+  masterWhere: TLBusWrapperLocation = CBUS // This needs to write to clint and bootaddrreg, which are on CBUS/PBUS
+)
+
+case object CustomBootPinKey extends Field[Option[CustomBootPinParams]](Some(CustomBootPinParams()))
+
+class WithCustomBootPinAltAddr(address: BigInt = 0x10040) extends Config((site, here, up) => {
```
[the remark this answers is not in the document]
jerryz123: Yeah good point

src/main/scala/BootAddrReg.scala
jerryz123: Probably no buffer is necessary.

src/main/scala/CustomBootPin.scala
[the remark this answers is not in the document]
jerryz123: I'm actually going to remove the default here. In practice you are always going to explicitly set this address to point to some other boot ROM/RAM specific to your setup.

jerryz123:
This has all been tested by me using these features to simulate various boot sequences.

src/main/scala/BootAddrReg.scala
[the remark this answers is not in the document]
abejgonzalez: Interesting. I didn't know that.

abejgonzalez (approved):
LGTM on the condition that you post an issue or track somewhere that we need to document all tapeout items (I'm pulling a bit of an Alon here).

jerryz123:
Issue for docs opened here: https://github.com/ucb-bar/chipyard/issues/844

colinschmidt (approved):
I'm ok with this. I'd prefer software for the boot pin detection but don't care that much given the small size of the state machine.

PR #129, opened 2021-03-31
Add TLROM which sits alongside the external serial memory

jerryz123:
This will go in after #128 .

This adds a TLROM which sits alongside the backing memory behind the TLSerial in the Harness. 
The intended purpose of this TLROM is as an alternate initial boot device that the BootAddrReg can select instead of DRAM_BASE. The default contents of the TLROM is just a `j DRAM_BASE`, but the intention is this will be replaced with chip-specific boot code. For example, code to setup chip SCRs, or code to copy binaries from sdcard into DRAM.

While the sdboot functionality can still be implemented as an alternative BootROM contents, having it live offchip offloads some verification work to post-silicon.

This has been tested in simulation. With the features in https://github.com/ucb-bar/testchipip/pull/128, I can redirect hart0 to jump to code in the TLROM after being woken by MSIP.


colinschmidt (commented):
I don't like having an address  <= DRAM_BASE off chip. It seems like this would complicate all on-chip routing but I'm not sure by how much.

jerryz123:
We already need to route the high bits to disambiguate between 0x8000_0000 and 0x9000_0000. I don't think adding a low address would increase the number of wires the serdes needs to consider? Unless there's something else I'm missing here.

Also the TLROM base is configurable. If it does become a problem, we can set it to a high address like 0x7000_0000.

src/main/scala/SerialAdapter.scala
```diff
@@ -462,7 +508,8 @@ class MultiClockSerialAXIRAM(
   }
 
   val memXbar = memClkRstDomain { TLXbar() }
-
+  val rom = memClkRstDomain { SerialTLROM(romParams, memParams.beatBytes) }
+  rom.node := TLFragmenter(memParams.beatBytes, p(CacheBlockBytes)) := memXbar
```
abejgonzalez, line 512: Shouldn't the `TLFrag` be in the `memClkRstDomain`?

src/main/scala/SerialAdapter.scala
```diff
@@ -303,8 +309,14 @@ case class AXIMemOverSerialTLClockParams(
     }
   }
 }
+case class SerialTLROMParams(
+  address: BigInt = 0x20000,
+  size: Int = 0x10000,
```
abejgonzalez, line 314: Can't you just infer the size based off the file / array being put into the ROM? That would minimize unused space.

src/main/resources/testchipip/tlrom/rom.S
```diff
@@ -0,0 +1,9 @@
+#define DRAM_BASE 0x80000000
+
+// Dummy empty alternate boot rom.
+// Immediately jumps to DRAM_BASE
+.section .text.start, "ax", @progbits
```
abejgonzalez, line 5: Can we just use this file as the default instead of hardcoding the array in the `SerialTLROM`? AKA compile this... commit it... and read in the file.

abejgonzalez (changes requested):

src/main/scala/SerialAdapter.scala
jerryz123: Yes. Good catch

abejgonzalez:
Im not sure I follow the issue with TLROM_ADDR <= DRAM_BASE... can you elaborate Colin.

src/main/scala/SerialAdapter.scala
jerryz123: The size is like a "max size". This is the same way it is done for the on-chip rom.
Also, since this is a ROM, having size > actual ROM contents size does not cost anything.

src/main/resources/testchipip/tlrom/rom.S
jerryz123: Reading in the file is difficult... For the bootrom we have to copy it to a top-level directory under Chipyard with GenerateSimFiles. For FireSim we have to use a separate WithBootrom fragment with firesim specific paths.

This default tlrom is practically useless. The intention is that you would replace this with your actual boot code post-silicon. 
I added the dummy `j DRAM_BASE` for testing, to confirm that the core can jump to here, execute code, and then jump into DRAM.

colinschmidt:
In general everything > DRAM_BASE is cacheable and everything less than it isn't. I think this will cause some extra gates in tilelink Xbars but its probably not a big deal. I also first thought the TLROM would not be cacheable and so different TL messages would be needed but it has the same type as DRAM so I think that makes the extra HW slightly less.
Either way it would end up being rather small if it exists at all and you wouldn't know for sure the cost unless you built the design.

colinschmidt (approved):

src/main/scala/SerialAdapter.scala
abejgonzalez: Yea I know. I have a similar issue with the original TLROM... I still think exposing a parameter that technically isn't necessary is useless. I won't die on this hill though.

abejgonzalez (approved):
LGTM

PR #130, opened 2021-04-05
Fix clock domain/device name of periphery scratchpad

jerryz123:


src/main/scala/Scratchpad.scala
```diff
@@ -20,7 +20,7 @@ trait CanHaveBackingScratchpad { this: BaseSubsystem =>
   private val portName = "Backing-Scratchpad"
 
   val spadOpt = p(BackingScratchpadKey).map { param =>
-    val spad = LazyModule(new TLRAM(address=AddressSet(param.base, param.mask), beatBytes=mbus.beatBytes))
+    val spad = mbus { LazyModule(new TLRAM(address=AddressSet(param.base, param.mask), beatBytes=mbus.beatBytes, devName=Some("scratchpad"))) }
```
colinschmidt, line 23: Can we call it `backingScratchpad` or something? It should be clear that this is a very special scratchpad not a generic RAM.

abejgonzalez (commented):
LGTM. Was this originally using the uncore top-level clock? Do all the adapters between the SPAD and the MBUS given by `toVariableWidthSlave` also get the MBUS clock (or the uncore clock)?

jerryz123:
The original used the uncore top-level clock.

`toVariableWidthSlave` puts the adapters in the mbus domain.

abejgonzalez (approved):

PR #132, opened 2021-05-04
add default member initializer for mm_dramsim2_t::clock_hz

timsnyder-siv:
in firesim/firesim#764 MIDAS-level simulation has been timing out.  This
is being caused by uninitialized value of clock_hz.  In general all
non-static data members should be given default member initializers
to guard against cases like this where several of the constructor
signatures of mm_dramsim2_t fail to initialize clock_hz.


jerryz123:
Thanks

jerryz123 (approved):

PR #136, opened 2021-09-13
Move ROM base to 0x10040

schwarz-em:
This change allows the bootrom to work for the [Ibex](https://github.com/lowRISC/ibex) core in Chipyard, since the Ibex reset vector is located at 0x10080.


jerryz123:
I think this change would affect all cores. Can we change reset vector of ibex instead

abejgonzalez:
> I think this change would affect all cores. Can we change reset vector of ibex instead

Unfortunately, we can't. Ibex uses a single boot address to determine where mtvec and the reset vector is. Thus BOOT_ADDR + 0x0 is the start of the vector table / mtvec and BOOT_ADDR + 0x80 is the reset vector.

Doc Ref: https://ibex-core.readthedocs.io/en/latest/03_reference/exception_interrupts.html?highlight=reset%20vector#exceptions-and-interrupts
Code: https://github.com/lowRISC/ibex/blob/6cbd7d21c541d7c849272d0cc490889bc4bf660c/rtl/ibex_if_stage.sv#L185

I think this change isn't too bad since we would just change the BootROM hang and address location to be 0x10080 and x10040 respectively for all CY cores (in the `WithBootROM` fragment). And even though mtvec is pre-set in Ibex, we override it in the bootrom.

jerryz123:
If we do want to change all cores (which I don't think we should do), then we should still leave bootrom base at 0x10000, and only the hang offset (equivalent to the reset vector) should be set to 0x80. The current changes change the base address, which is strange to me.

I think a better solution is to modify the bootrom to simultaneously support cores that reset to 0x10040 and 0x10080. This should be straightforward if you make the instruction at 0x10080 a `j 0x10040`.

src/main/resources/testchipip/bootrom/bootrom.S
```diff
@@ -37,6 +37,15 @@ wfi_loop: // wait for MSIP interrupt to start program execution
   wfi
   j wfi_loop
 
+// jump to reset vector
+.section .text.hang80, "ax", @progbits
+.globl _hang80
+_hang80:
+  j _hang
+
+.section .text.hangC0, "ax", @progbits
```
abejgonzalez, line 46: I think you don't need this other section? If you remove it, the address of the `boot_core` section will be immediately after the `j _hang` instruction... which is fine.

schwarz-em: You're definitely right... I'll fix that.

abejgonzalez: You should recompile the bootrom images and rebump the CY PR as well.

abejgonzalez (approved):
LGTM.

PR #141, opened 2022-01-28
Support multi-thread VCS simv option like FGP, Xprof etc.

SihaoLiu:
According to https://github.com/ucb-bar/chipyard/issues/1081 and per @jerryz123 's suggestion, I remove the VCS simv option from argv before entering base htif constructor. It is tested under chipyard. 

Command like `make EXTRA_SIM_FLAGS="-fgp=num_threads:2 -Xdprof=timeline"` will enable multi-thread VCS simulation via two step flow:
```
% vcs -fgp -full64 <otherOptions> 
% simv -fgp=num_threads:<value>
```

Additionally, a minor fix mentioned https://github.com/ucb-bar/testchipip/pull/134 due to compilation error.

### DIFF
--- a/src/main/resources/testchipip/csrc/SimSerial.cc
+++ b/src/main/resources/testchipip/csrc/SimSerial.cc
@@ -6,6 +6,24 @@
 
 testchip_tsi_t *tsi = NULL;
 
+// Remove VCS simv option from argv if it match pattern -X???=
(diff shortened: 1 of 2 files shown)


harrisonliew (approved):
Tested this out, LGTM.

PR #143, opened 2022-03-06
Idle tsi in the target thread.

jerryz123:
Switch to the target thread when `htif` `idle`s. This enables continual execution of the target even if htif isn't working correctly.


jerryz123:
Fixes https://github.com/ucb-bar/chipyard/issues/1136

harrisonliew (approved):
Didn't test functionality but looks correct to me and compiled. CI failure also seems irrelevant (riscv-tools timeout).

PR #146, opened 2022-05-16
[WIP] Add PTWProber

jerryz123:
This is a RoCC "accelerator" that makes it fast and easy for user programs to determine page mappings. It supports a single instruction that performs a PTW using the PTW on the host core.

```
#include <stdio.h>
#include <stdint.h>
#include "rocc.h"

int main(void)
{
  int stack_val = 0;

  uint64_t vaddr = (uint64_t)&stack_val;
  uint64_t paddr;
  // opcode=0
  ROCC_INSTRUCTION_DS(0, paddr, vaddr, 0)

  printf("%lx to %lx\n", vaddr, paddr);
  return 0;
}
```

### DIFF
--- a/src/main/scala/PTWProber.scala
+++ b/src/main/scala/PTWProber.scala
@@ -1,0 +1,55 @@
+package testchipip
+
+import chisel3._
+import chisel3.util._
+
+import freechips.rocketchip.tile._
+import freechips.rocketchip.rocket._
+import freechips.rocketchip.config.{Parameters}
+
+/*
+ * RoCC "accelerator" which translates a virtual address
+ * Useful for giving user programs info about physical addresses
+ */
+class PTWProber(opcodes: OpcodeSet)(implicit p: Parameters) extends LazyRoCC(opcodes, 1) {
+  override lazy val module = new LazyRoCCModuleImp(this) with HasCoreParameters {
+    val in_q = Queue(io.cmd)
+    val in_flight = RegInit(false.B)
+    val resp_q = Module(new Queue(new RoCCResponse, 2))
+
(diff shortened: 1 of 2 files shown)


src/main/scala/PTWProber.scala
```diff
@@ -0,0 +1,55 @@
...
+    } .elsewhen (io.ptw(0).resp.valid) {
+      in_flight := false.B
+    }
+    in_q.ready := Mux(in_flight, io.ptw(0).resp.valid, !vm_enabled && resp_q.io.enq.ready)
+
+
+    resp_q.io.enq.valid := Mux(in_flight, io.ptw(0).resp.valid, !vm_enabled && in_q.valid)
+    resp_q.io.enq.bits.rd := in_q.bits.inst.rd
+    resp_q.io.enq.bits.data := Mux(in_flight,
+      Mux(io.ptw(0).resp.bits.pf, 0.U, Cat(io.ptw(0).resp.bits.pte.ppn, in_q.bits.rs1(pgIdxBits-1,0))),
```
a0u, line 40: Is the intent here to determine whether a mapping exists or whether an access would succeed?

As far as I know, `io.ptw(0).resp.bits.pf` only indicates if the PTE is invalid or the reserved bits are non-zero, not necessarily if the permissions are adequate.  Of course, this suffices for the first use case but not the latter.

Access exceptions (`ae_ptw` and `ae_final`) should probably also mask the result.

a0u (commented):
A very unlikely race condition could occur where the core takes an exception after the RoCC instruction is issued and privileged code modifies the `satp` CSR before the walk completes.  Not much could be done about this, though, with the RoCC interface being asynchronous and no way to provide a separate copy of `satp` to the PTW without changing rocket-chip.

PR #152, opened 2022-12-31
Updating build.sbt as a precursor to updating Maven

tianrui-wei:


tianrui-wei:
This commit bumps `build.sbt` in preperation for bumping the published version on maven, which is ~3 years old at this point.

jerryz123:
I feel like this will break when testchipip is built with chipyard, since the chipyard testchipip project will inherit these settings.

jerryz123:
The dependency on sifive-blocks is regrettable. IMO we should refactor the stuff that depends on sifive-blocks out of this into chipyard directly.

UARTAdapter and SPIFlash can both go in chipyard. Then testchipip only needs rocket-chip

tianrui-wei:
I feel chipyard should override the dependency in its settings when you flip on a development switch, in which case it will clone dependencies as submodules. But in default mode, it should just use binary dependency off maven.

Until we figure out how to publish CY as a maven package, I'd avoid moving stuff into chipyard, as the original intention would be enabling stand-alone use of packages, rather than have everything depend on chipyard.

jerryz123:
> I feel chipyard should override the dependency in its settings when you flip on a development switch, in which case it will clone dependencies as submodules. But in default mode, it should just use binary dependency off maven.

Ok, I think that might be ok for now, but I think we should avoid adding more overrides to chipyard's build.sbt.

tianrui-wei:
Sorry for the blip, restored the old master branch

jerryz123:
I meant open a PR for chipyard, then we merge that PR and this one at the same time, once we confirm the chipyard changes work. 

Its not a big deal, but I stlil reverted your changes for now, also so we can get feedback from @abejgonzalez and others

tianrui-wei:
Sounds good, yeah sorry for the wrong push. I was meaning to push to my own fork... Thanks for the help.

Actually this depends on https://github.com/chipsalliance/rocket-chip-blocks/pull/9 so this should probably not be merged until that one is merged and a maven package is already upstream

tianrui-wei:
CY PR opened at https://github.com/ucb-bar/chipyard/pull/1299, new PR at https://github.com/ucb-bar/testchipip/pull/153

jerryz123:
For Chipyard I mean a chipyard PR that bumps testchipip submodule, and fixes and issues with build.sbt.

tianrui-wei:
Oh I see lol

PR #154, opened 2023-01-04
[WIP] Bump to scala 2.13/chisel 3.5.5/latest rocketchip

jerryz123:


src/main/scala/Serdes.scala
```diff
@@ -555,7 +556,8 @@ class TLDesser(w: Int, params: Seq[TLClientParameters], hasCorruptDenied: Boolea
   val node = TLClientNode(params.map(client =>
       TLMasterPortParameters.v1(Seq(client))))
 
-  lazy val module = new LazyModuleImp(this) {
+  lazy val module = new Impl
+  class Impl extends LazyModuleImp(this) {
```
abejgonzalez, line 560: Why is this needed/done?

abejgonzalez (approved):

src/main/scala/TraceIO.scala
```diff
@@ -13,45 +13,17 @@ import freechips.rocketchip.tile.{BaseTile}
 import freechips.rocketchip.diplomacy.{BundleBridgeSource, BundleBroadcast, BundleBridgeNexusNode}
 
 //***************************************************************************
-// Extended Trace Instruction Utilities:
-// Used to map TracedInstructions to their Extended version and
```
abejgonzalez, line 17: Why is this getting removed?

jerryz123: I upstreamed this to rocket-chip

src/main/scala/Serdes.scala
jerryz123: Result of change in scala2.13. This change was applied via scalafix, and all rocket-chip dependencies on 2.13 had this change applied

abejgonzalez (approved):

PR #164, opened 2023-04-09
feat: determinstic initialization for mm

tianrui-wei:
This commit adds memory poisoning mechanism

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

### DIFF
--- a/src/main/resources/testchipip/csrc/mm.cc
+++ b/src/main/resources/testchipip/csrc/mm.cc
@@ -6,7 +6,12 @@
 #include <cstdlib>
 #include <cstring>
 #include <cassert>
+#include <chrono>
 #include <sys/mman.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/syscall.h>
 
 void mm_t::write(uint64_t addr, uint8_t *data, uint64_t strb, uint64_t size)
 {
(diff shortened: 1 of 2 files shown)


src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -38,19 +43,36 @@ void mm_t::init(size_t sz, int wsz, int lsz)
   assert(wsz > 0 && lsz > 0 && (lsz & (lsz-1)) == 0 && lsz % wsz == 0);
   word_size = wsz;
   line_size = lsz;
+  seed = std::chrono::system_clock::now().time_since_epoch().count();
```
jerryz123, line 46: Lets make this seed a static constant. Otherwise behavior will change across compilations of the same config, which complicates debugging

src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -38,19 +43,36 @@ void mm_t::init(size_t sz, int wsz, int lsz)
...
+  std::cout << "didn't supply a seed! Random seed is " << seed << std::endl;
```
jerryz123, line 47: Can you make this more descriptive? Ex:
"DRAMSim mm_t init seed is ..."

src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -38,19 +43,36 @@ void mm_t::init(size_t sz, int wsz, int lsz)
...
+  mt = std::mt19937(seed);
+  std::string path_name = "chipyard-cosim-" + std::to_string(getpid());
```
jerryz123, line 49: The fd for the shared memory object should not contain references to cosim. Name this something like `mm_<pid>`

src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -38,19 +43,36 @@ void mm_t::init(size_t sz, int wsz, int lsz)
...
-          NULL, sz, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, -1, 0);
+          NULL, sz, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, shared_fd, 0);
   if (data == MAP_FAILED) {
     std::perror("[mm_t] mmap for backing storage failed");
     exit(-1);
   }
 
   size = sz;
+  uint64_t* data_64 = (uint64_t *)data;
+  for (uint64_t i = 0; i < (size / 8); i++) {
```
jerryz123, line 65: Can you add an assert here that `size % 8 == 0`

src/main/resources/testchipip/csrc/mm.cc
```diff
@@ -38,19 +43,36 @@ void mm_t::init(size_t sz, int wsz, int lsz)
...
+	  data_64[i] = mt();
+  }
+  printf("mm: finished initialization, thread is is %lx, process id is %lx, parent p id is %lx\n", syscall(__NR_gettid),getpid(), getppid());
```
jerryz123, line 68: Can you fix the indentation in this file to be space-based only?

jerryz123 (changes requested):

PR #168, opened 2023-04-18
Flip serialized tilelink clock source

jerryz123:
Generate the clock off-chip, instead of relying on the on-chip clock

### DIFF
--- a/src/main/scala/SerialAdapter.scala
+++ b/src/main/scala/SerialAdapter.scala
@@ -21,29 +21,6 @@ import java.nio.file.{Files, Paths}
 case object SerialAdapter {
   val SERIAL_TSI_WIDTH = 32 // hardcoded in FESVR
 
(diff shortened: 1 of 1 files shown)


harrisonliew (changes requested):
SimSerial needs to have its clock direction flipped too, right?
Also comment about using serial_tl_domain's reset seems reasonable - it get's sync'd over the clock crossing?

jerryz123:
This shouldn't affect SimSerial, since that is a simulation-only construct. This only affects the serialized tilelink interface between the harness (or FPGA) and the chip.

jerryz123:
Hmm, I think the reset on the external side needs to be sync'd

harrisonliew:
> This only affects the serialized tilelink interface between the harness (or FPGA) and the chip.

When generating the harness/FPGA interface, how does its clock get flipped properly? Do the XDC constraints need to be updated too?

T-K-233 (approved):
✓

PR #170, opened 2023-04-21
FIX (improve-boot-1): change reset vector address and boot register address

T-K-233:
Change reset vector PC address to 0x10000. This makes more sense, and users can override BOOTROM with their own code when testing.

### DIFF
--- a/src/main/scala/Serdes.scala
+++ b/src/main/scala/Serdes.scala
@@ -594,10 +594,20 @@ class TLDesser(w: Int, params: Seq[TLClientParameters], hasCorruptDenied: Boolea
   }
 }
 
+object TLSerdesser {
+  // This should be the standard bundle type for TLSerdesser
+  val STANDARD_TLBUNDLE_PARAMS = TLBundleParameters(
+    addressBits=64, dataBits=64,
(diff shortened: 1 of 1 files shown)


T-K-233:
Note: this depends on [rocket-chip #3328](https://github.com/chipsalliance/rocket-chip/pull/3328)

jerryz123:
This should not require rocketchip changes. In Chipyard you can change the withbootrom config fragment to also set hang

tianrui-wei:
Note that this PR can be potentially disruptive as we are trying to bring in force support

jerryz123:
From a separate discussion, I think we will not change the default bootrom. Instead we will improve support for custom next stage bootloaders.

T-K-233:
> Note that this PR can be potentially disruptive as we are trying to bring in force support

Sorry I don't quite understand, what do you mean by "bring in force support"...?

tianrui-wei:
Sorry for the confusion. For context, we're using force-riscv in another branch. This PR looks very good and I'm fine with it being merged as long as Jerry approves it. This is more of a note to Jerry as we try to upstream things.

PR #172, opened 2023-05-04
Fix Multi-UART | add ForcePTY

jerryz123:
- Fixes multi-uart designs, now the multiple uarts are independent, while previously, they were all multiplexed off the same thing
- Allows forcePTY, which forces a SimUART to send the UART out to a pty

Note that this does not require a Chipyard bump, since APIs are unchanged. To avoid more churn in chipyard, I'll just open the PR here, and the next testchipip bump in chipyard will pick up these changes too.


joey0320 (approved):
LGTM

PR #173, opened 2023-05-08
Make BootAddrReg an optionally-enabled feature

jerryz123:
Previously, BootAddrReg was always present on all designs.
It doesn't make since to add this to designs with no cores, so make this an optional device intsead.


T-K-233 (approved):
Looks good!

PR #175, opened 2023-05-10
Tracedoctor

bgottschall:
TraceDoctor support in Firesim.

These changes provide necessary TraceDoctorIO classes similar to TracerV. TraceDoctor also attaches here to the TracerV trigger system.


jerryz123:
We've upstreamed support into the existing TraceIO to allow that IO to carry an arbitrary payload alongside the instruction trace, we should movetowards upstreaming TraceDoctor using that interface.

bgottschall:
Awesome thanks! I have to see when I find some time to merge that into TraceDoctor.

PR #184, opened 2023-05-29
Improvements to clock dividers

jerryz123:
`ClockDivider` can be parameterized to reset with a active clock out, instead of gating it. 

`ClockDividerOrPass` now is a RawModule, so users have to be explicit in specifying what clock is being divided

`io.divisor` in `ClockDividerOrPass` is now sync'd to the clock it is dividing, as it should be.

### DIFF
--- a/src/main/scala/ClockUtil.scala
+++ b/src/main/scala/ClockUtil.scala
@@ -5,7 +5,7 @@ import chisel3.util._
 import chisel3.experimental.{Analog, IntParam}
 
 import org.chipsalliance.cde.config.Parameters
-import freechips.rocketchip.util.{AsyncResetSynchronizerShiftReg, ClockGate, ClockGateImpl}
+import freechips.rocketchip.util._
 
 class ClockFlop extends BlackBox with HasBlackBoxResource {
     val io = IO(new Bundle {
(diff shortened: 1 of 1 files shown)


src/main/scala/ClockUtil.scala
```diff
@@ -165,21 +165,21 @@ class ClockDivider(width: Int) extends Module {
...
+    val divisor = Input(UInt(width.W)) // this may not be synchronous with clockIn
     val resetAsync = Input(AsyncReset())
     val clockOut = Output(Clock())
 
   })
-
-  val divider = Module(new ClockDivider(width))
-  divider.io.divisor := io.divisor
+  val divider = withClockAndReset(io.clockIn, io.resetAsync) { Module(new ClockDivider(width, initDiv=1)) }
+  divider.io.divisor := withClockAndReset(io.clockIn, io.resetAsync) { SynchronizerShiftReg(Mux(io.divisor === 0.U, 1.U, io.divisor)) }
```
harrisonliew, line 177: The Mux seems unnecessary here. We can just gate the divider since the clock mux below is passing the clock thru when divisor === 0.U.

harrisonliew (approved):

src/main/scala/ClockUtil.scala
jerryz123: Unfortunately not, the `ClockMutexMux` needs the clocks to be alive, else it can get stuck waiting on a dead clock that never advances its synchronizer-shift-reg..

PR #196, opened 2023-09-05
Refactor: Adjust bootrom reset vector address

T-K-233:
Change corresponding to [chipyard #1534](https://github.com/ucb-bar/chipyard/pull/1534)

The motivation is that the current organization of the bootrom code and the reset vector is rather confusing. One would normally assume that the reset vector is pointed toward the base of a memory section (e.g. 0x10000).

This could also benefit future bootrom code development.


src/main/resources/testchipip/bootrom/bootrom.S
```diff
@@ -1,4 +1,4 @@
-#define BOOTADDR_REG 0x4000
+#define BOOTADDR_REG 0x1000
```
jerryz123, line 1: To be clear, this isn't related/required for the bootrom address changes right?

T-K-233: Yes, this is not related to the bootrom address change.

This change is corresponding to the proposal mentioned [here](https://docs.google.com/presentation/d/18qrFsHImYO4OJEpl8oQ_g3m2dc6a2Q0LppnVxTX-19I/edit#slide=id.g1c27f282ad8_0_54).

src/main/scala/BootAddrReg.scala
```diff
@@ -15,7 +15,7 @@ import freechips.rocketchip.prci._
 
 case class BootAddrRegParams(
   defaultBootAddress: BigInt = 0x80000000L, // This should be DRAM_BASE
-  bootRegAddress: BigInt = 0x4000,
+  bootRegAddress: BigInt = 0x1000,
```
jerryz123, line 18: If you make this change, you have to change here as well: https://github.com/ucb-bar/testchipip/blob/307fca3c3ec4484d071744df6752bbbb77be7ddb/src/main/resources/testchipip/csrc/cospike_impl.cc#L198

jerryz123 (approved):
Good to go

PR #201, opened 2023-10-06
Support sub-banking spads

jerryz123:


SeahK:
it worked, but didn't really improve performance

PR #210, opened 2023-12-19
Organize TestChipIP into sub-packages

jerryz123:
This makes it a bit easier to discover parts of testchipip.

The README is also updated.


src/main/scala/serdes/.#Adapter.scala
abejgonzalez, line 1: Extra file.

src/main/scala/serdes/Adapter.scala
```diff
@@ -0,0 +1,101 @@
...
+  val aggregator = Module(new SerialWidthAggregator(narrowW, wideW))
+  aggregator.io.narrow <> io.narrow.in
+  io.wide.out <> aggregator.io.wide
+
+  val slicer = Module(new SerialWidthSlicer(narrowW, wideW))
+  slicer.io.wide <> io.wide.in
+  io.narrow.out <> slicer.io.narrow
+}
+
+class UARTToSerial(freqHz: BigInt, uartParams: UARTParams) extends Module {
```
abejgonzalez, line 67: Duplicate? There is an `UARTToSerial` file anyways.

abejgonzalez (changes requested):

jerryz123:
@kenhoberkeley @T-K-233 any thoughts? Else I will merge

T-K-233 (approved):

PR #212, opened 2023-12-31
Add Chip ID Pin

schwarz-em:
Adds a pin to set the chip id for each chip. The pin drives a read-only MMIO register attached to the CBUS.

The purpose of adding this pin is to enable cache coherency across multi-chip configs in Chipyard.

### DIFF
--- a/src/main/scala/soc/ChipIdPin.scala
+++ b/src/main/scala/soc/ChipIdPin.scala
@@ -9,8 +9,8 @@ import freechips.rocketchip.regmapper._
 import freechips.rocketchip.subsystem._
 
 case class ChipIdPinParams (
-  numChips: Int = 1, 
-  chipIdAddr: BigInt = 0x2000, //TODO: pick addr
+  width: Int = 1, 
+  chipIdAddr: BigInt = 0x2000,
   slaveWhere: TLBusWrapperLocation = CBUS
 )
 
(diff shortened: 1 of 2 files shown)


src/main/scala/soc/ChipIdPin.scala
```diff
@@ -0,0 +1,44 @@
...
+import chisel3._
+import chisel3.util._
+import org.chipsalliance.cde.config._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.regmapper._
+import freechips.rocketchip.subsystem._
+
+case class ChipIdPinParams (
+  numChips: Int = 1,
```
jerryz123, line 12: Instead of `numChips`, lets just make this `width`. Everything is in relation to the width.

src/main/scala/soc/ChipIdPin.scala
```diff
@@ -0,0 +1,44 @@
...
+
+trait CanHavePeripheryChipIdPin { this: BaseSubsystem =>
+  val chip_id_pin = p(ChipIdPinKey).map { params => 
+    val tlbus = locateTLBusWrapper(params.slaveWhere)
+    val device = new SimpleDevice("chip-id-reg", Nil)
+    val id_bits = log2Up(params.numChips)
+
+    val inner_io = tlbus {
+      val node = TLRegisterNode(
+        address = Seq(AddressSet(params.chipIdAddr, 0xfff)), //TODO: fix address set size
```
jerryz123, line 27: Lets make the address set size 4096 bytes, to match all the other devices.

A wider size than the actual register also simplifies address decoding.

jerryz123 (commented):
Minor nits, otherwise looks good here

jerryz123 (approved):

PR #214, opened 2024-01-03
Update deprecated APIs to prepare for Chisel 5

tymcauley:
- `IO` was moved from `chisel3.experimental` to `chisel3`
- Calls to `fire` with empty parens need the parens removed
- `DataMirror` was moved from `chisel3.experimental` to `chisel3.reflect`


jerryz123:
Does this break with chisel3.6? We need it to work just a little bit longer on 3.6

jerryz123 (approved):

PR #216, opened 2024-01-23
Add flag to disable cospike instruction trace printf

jerryz123:


joey0320 (commented):
maybe we change `if (cospike_printf)`  to `if (unlikely(cospike_printf))` for slight perf?

joey0320:
Otherwise LGTM

joey0320 (approved):

jerryz123:
Any perf benefits from that would be minimal due to amdahl's

PR #218, opened 2024-01-29
Refactor/fix/improve the TileLink serializer

jerryz123:
This refactor of the TL serializer improves performance and fixes deadlock-conditions.

- A TileLink message is serialized into compacted-beats, where the head beat contains the constant fields (address/opcode/id), while the body beats carry the variable fields (data/mask). A n-beat TileLink burst gets transformed into n+1 compacted beats. This avoids sending redundant address/opcode/id bits over the wire
- The TileLink compacted-beat format is serialized per-channel into Flits of configurable width.
- A "SerialPhy" layer is constructed between the Serdesser and the chip I/O, this layer implements clock crossings and flow control
- The Phy layer breaks the Flits into configurable-width Phits. phitWidth determines the number of I/O pins necessary
- For all Phy types, non-blocking muxing between the channels is provided, so channel deadlock is not possible
- For the Credited flow-control type, credits and data share the same Flit/Phit interface

Note: This serializer is not backwards compatible with the old implementation

This PR also removes `tsiHost`, as now we directly support FESVR-host-on-PC bringup (instead of FESVR-host-on-FPGA).


vikramjain236 (approved):

abejgonzalez (approved):
SFTM

PR #231, opened 2024-05-30
Set names of generated clock domains

jerryz123:


kevindna:
does this only affect the scratchpad and the serial TL? in other words, are _all_ the clockdomains given apropriate names

jerryz123:
No. This only affects the fixes for those blocks. Fixes for the other unnamed ClockDomains are inflight in the other repos

kevindna:
incremental progress is still progress.

kevindna (approved):

PR #239, opened 2024-07-18
Clear MIP in BootROM

jerryz123:
I'll bump CY with this to check with CY's CI


T-K-233 (approved):
LFTM

abejgonzalez (approved):
Lgtm. For my understanding, why is this needed?

jerryz123:
The MIP CSR is not reset-iniitalized, which means it can take on random values between simulation runs. While correctly written M-mode firmware handles this with no problem, it does cause some minor divergence in code path. 

This non-determinism is undesirable.

PR #246, opened 2024-08-16
Improve testchip_uart_tsi

jerryz123:
* Binary loading is now bursty, instead of buffering all the bytes in memory before sending any, it streams chunks into the targe
 * Self-check is now per-chunk, instead of once at the end of the binary

### DIFF
--- a/uart_tsi/testchip_uart_tsi.cc
+++ b/uart_tsi/testchip_uart_tsi.cc
@@ -7,6 +7,7 @@
 #include <errno.h>
 #include <termios.h>
 
+#define PRINTF(...) printf("UART-TSI: " __VA_ARGS__);
 
 testchip_uart_tsi_t::testchip_uart_tsi_t(int argc, char** argv,
 					 char* ttyfile, uint64_t baud_rate,
@@ -37,17 +38,17 @@ testchip_uart_tsi_t::testchip_uart_tsi_t(int argc, char** argv,
   case 3000000: baud_sel = B3000000; break;
   case 4000000: baud_sel = B4000000; break;
   default:
-    printf("Unsupported baud rate %ld\n", baud_rate);
+    PRINTF("Unsupported baud rate %ld\n", baud_rate);
     exit(1);
   }
 
   if (baud_sel != B115200) {
-    printf("Warning: You selected a non-standard baudrate. This will only work if the HW was configured with this baud-rate\n");
+    PRINTF("Warning: You selected a non-standard baudrate. This will only work if the HW was configured with this baud-rate\n");
   }
 
   ttyfd = open(ttyfile, O_RDWR);
   if (ttyfd < 0) {
-    printf("Error %i from open: %s\n", errno, strerror(errno));
+    PRINTF("Error %i from open: %s\n", errno, strerror(errno));
     exit(1);
   }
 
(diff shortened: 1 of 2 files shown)


T-K-233 (approved):
Looks fine

uart_tsi/testchip_uart_tsi.cc
```diff
@@ -142,62 +143,50 @@ bool testchip_uart_tsi_t::check_connection() {
...
-	printf("Conflict addr: %lx - %lx\n", it.first, it.first + it.second.size());
-	exit(1);
+  while (this->handle_uart()) { }
+  if (this->in_load_program) { printf("Done\n"); }
+
+  if (this->do_self_check && this->in_load_program) {
+    uint8_t rbuf[chunk_max_size()];
+    const uint8_t* csrc = (const uint8_t*)src;
+    PRINTF("Performing self check of region %lx-%lx ... ", taddr, taddr + nbytes);
+    read_chunk(taddr, nbytes, rbuf);
```
vighneshiyer, line 170: `read_chunk` seems to delegate to `tsi_t::read_chunk` which is a blocking call effectively (https://github.com/riscv-software-src/riscv-isa-sim/blob/master/fesvr/tsi.cc#L52). Why is the `while (handle_uart()) {}` needed below?

vighneshiyer (commented):
lgtm, I just had a few questions

uart_tsi/testchip_uart_tsi.cc
```diff
@@ -142,62 +143,50 @@ bool testchip_uart_tsi_t::check_connection() {
...
-      }
-      printf("Self check succeeded chunk %lx to %lx\n", addr, addr + it.second.size());
-    }
-    printf("Self check success\n");
-  }
 }
 
 void testchip_uart_tsi_t::write_chunk(addr_t taddr, size_t nbytes, const void* src) {
+  if (this->in_load_program) { PRINTF("Loading ELF %lx-%lx ... ", taddr, taddr + nbytes); }
   testchip_tsi_t::write_chunk(taddr, nbytes, src);
```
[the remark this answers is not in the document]
jerryz123: I wouldn't say any of this is a "bug fix", as I could not determine if there was any bug in the original implementation. I just found the original approach might result in unintuitive (but technically correct) behaviors.

The prior approach could end up in a situation where the entire ELF Is loaded into a byte-stream buffer in memory (Which could be very large), before `handle_uart` goes through the process of actually streaming it to the chip. Additionally, selfcheck would only be run after the entire program is loaded.

The loaded_program map was to remember the memory layout of the program such that selfcheck could work. However, this ran into problems when the ELF encodes overlapping memory regions (load_program might write the same memory region twice with different values, complicating selfcheck).

Now self-check is done on a chunk-by-chunk basis. Not only does this give you faster feedback when you write into a bad memory region, it also throttles transmit, decreasing the probability of inducing a irrecoverable backpressure condition on the FPGA (although this should already be highly unlikely).

uart_tsi/testchip_uart_tsi.cc
jerryz123: Ah I think that one is unnecessary. But it also shouldn't hurt.

PR #247, opened 2024-09-30
Improve cospike API for multiple memory regions

jerryz123:
This doesn't remove the legacy `mem0_base/size` field, so firesim+cosim will still work correctly.
At some point the next user of firesim+cosim may want to update the bridge to use the new API, but that can be done in the future.

### DIFF
--- a/src/main/scala/cosim/Cospike.scala
+++ b/src/main/scala/cosim/Cospike.scala
@@ -14,15 +14,17 @@ case class SpikeCosimConfig(
   priv: String,
   pmpregions: Int,
   maxpglevels: Int,
-  mem0_base: BigInt,
-  mem0_size: BigInt,
   nharts: Int,
   bootrom: String,
   has_dtm: Boolean,
+  mems: Seq[(BigInt, BigInt)],
+  // Legacy APIs
+  mem0_base: BigInt = 0,
+  mem0_size: BigInt = 0,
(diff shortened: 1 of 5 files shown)


src/main/scala/cosim/Cospike.scala
```diff
@@ -64,10 +66,20 @@ class SpikeCosim(cfg: SpikeCosimConfig) extends BlackBox(Map(
...
+{
+  val io = IO(new Bundle {})
+}
+
 object SpikeCosim
 {
   def apply(trace: TileTraceIO, hartid: Int, cfg: SpikeCosimConfig) = {
     val cosim = Module(new SpikeCosim(cfg))
+    for ((base, size) <- cfg.mems) {
+      val reg = Module(new SpikeCosimRegisterMemory(base, size))
```
abejgonzalez, line 81: Is this module's DPI calls in the `initial` block guaranteed to be before the `SpikeCosim` module in this code? I'm not sure what order initial blocks are called if the `reg` module is after the `cosim` module in the code.

jerryz123: The `initial` blocks can be called in any order, but all the initial blocks should be executed before any of the actual simulation verilog is evaluated.

abejgonzalez (approved):

PR #248, opened 2024-10-17
Misc Xcelium fixes

abejgonzalez:
Masks of `mem_base` in `SimDRAM.cc` due to a sign extension bug. Also adds a default parameter value for `SimTSI`.

### DIFF
--- a/src/main/resources/testchipip/csrc/SimDRAM.cc
+++ b/src/main/resources/testchipip/csrc/SimDRAM.cc
@@ -22,9 +22,15 @@ extern "C" void *memory_init(
                              long long int line_size,
                              long long int id_bits,
                              long long int clock_hz,
-                             long long int mem_base
+                             long long int mem_base,
+                             int addr_bits
 			     )
 {
+    // in xcelium, while mem_base is a 40b unsigned param, when passed into a long long int DPI variable it becomes signed.
+    // mask off the upper address bits (since we enforce that only address bits worth of address space is available).
+    uint64_t mem_base_mask = (1ULL << addr_bits) - 1;
+    mem_base &= mem_base_mask;
+
     mm_t *mm;
     s_vpi_vlog_info info;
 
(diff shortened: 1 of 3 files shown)


src/main/resources/testchipip/csrc/SimDRAM.cc
```diff
@@ -25,6 +25,14 @@ extern "C" void *memory_init(
                              long long int mem_base
 			     )
 {
+    mem_size &= 0xFFFFFFFF;
+    word_size &= 0xFFFFFFFF;
+    line_size &= 0xFFFFFFFF;
+    id_bits &= 0xFFFFFFFF;
+    clock_hz &= 0xFFFFFFFF;
+    mem_base &= 0xFFFFFFFF;
```
abejgonzalez, line 33: Maybe we should mask based on the ADDR_SIZE parameter (pass that into these functions and use it to mask at least mem_base)?

src/main/resources/testchipip/vsrc/SimTSI.v
```diff
@@ -10,7 +10,7 @@ import "DPI-C" function int tsi_tick
     output int tsi_in_bits
 );
 
-module SimTSI #(parameter CHIPID) (
+module SimTSI #(parameter CHIPID=0) (
```
abejgonzalez, line 13: Needed for xcelium to not error.

src/main/resources/testchipip/csrc/SimDRAM.cc
abejgonzalez: Any thoughts on my mem_base might be sign extended in this function (the parameter in `SimDRAM.v` is a 40b `0x80000000` but when you print it's output here it is `0xFFFF8....`).

jerryz123:
Are there any differences in C compiler flags between the two sims?

abejgonzalez:
Nope. This is some difference in how DPI works across simulators. Unfortunately, it looks like the behavior between all 3 simulators is different (I've experienced differences between VCS/Verilator... and now VCS/Xcelium).

jerryz123 (approved):

PR #251, opened 2025-02-05
Bootrom and DRAM changes for Radiance

richardyrh:
* adds the assembly, binary and dump for radiance's boot rom
* parameterizes DRAM width so that reads and writes can be more than 64 bits, should not affect existing designs


jerryz123 (approved):

jerryz123 (commented):
The DRAM stuff is good. Are you sure you want to put the bootrom in here? It might make it easier for you to leave it in the Radiance repo, since you can iterate on it without having to touch testchipip.

If you think it should be in here, I have no problem with merging it.

richardyrh:
The bootrom hasn't been touched in a while, so should be stable. Plus I think the build flow is easier if it's in testchipip. (sorry for merging already)

PR #255, opened 2025-04-07
Add SubsystemInjector API to reduce dependency on cake pattern

jerryz123:


iansseijelly:
Looks good to me.

iansseijelly (approved):
LGTM

PR #263, opened 2025-11-24
Add CTC

schwarz-em:
Same as Lux's original CTC PR but with new features:

- Allow removing the PHY for the FireSim bridge
- Support multiple CTC ports per config
- Adds the standalone address translator (finally)


iansseijelly (approved):
LGTM

PR #265, opened 2025-12-11
Fix CTC unit test

schwarz-em:
Modify the CTC unittest so that it only sends requests >=32 bits, as CTC cannot handle sub-word requests.

Also, add CTC to the readme


iansseijelly (approved):
LGTM

PR #269, opened 2026-01-29
Bidirectional Address Translation and CTC Debug Collateral

schwarz-em:
The purpose of this PR is to enable building a "passthrough" IO chiplet within Chipyard

It contributes:
- Target-side address translation (controlled by the type of params)
- An offchip test memory like SerialRAM but for CTC, which enables modeling other chiplets as memory
- Support for a decoupled PHY in CTC
- A bugfix to CTC for incorrect TL beat handling on the tl2ctc side


iansseijelly (approved):
lgtm

